PR #17, opened 2018-03-30
eos-paygd: Fix meson conflation between eos-paygd.conf output files

pwithnall:
**This only needs to be merged on eos3.3.**

If meson has two configure_file() rules with the same output in the same
directory, one of them will silently overwrite the other. This was
causing the sysusers.d rules to be installed in the tmpfiles.d directory
as well as the sysusers.d directory, and causing tmpfiles.d to fail.

Fix that by renaming the eos-paygd.conf for tmpfiles.d to
eos-paygd-state.conf, taking advantage of tmpfiles.d’s support for
‘parts’ of packages.

Signed-off-by: Philip Withnall <withnall@endlessm.com>

https://phabricator.endlessm.com/T21730


pwithnall:
I looked at filing this as an upstream issue, but it looks like it’s covered by these two already:
 * https://github.com/mesonbuild/meson/issues/2320
 * https://github.com/mesonbuild/meson/issues/2634

PR #19, opened 2018-04-12
libeos-payg: Poll for the expiry time to avoid issues with suspend

pwithnall:
Rather than setting a single long timeout period and then expiring when
that is reached, poll every 60s to see if the expiry time has been
reached. This avoids the timeout going out of sync with the wall clock
when resuming from suspend, and also works around issues with the system
clock and timezone changing.

Signed-off-by: Philip Withnall <withnall@endlessm.com>

https://phabricator.endlessm.com/T22074

### DIFF
--- a/libeos-payg/manager.c
+++ b/libeos-payg/manager.c
@@ -494,14 +494,23 @@ check_is_counter_unused (EpgManager  *self,
 }
 
 static gboolean
-expired_cb (gpointer user_data)
+check_expired_cb (gpointer user_data)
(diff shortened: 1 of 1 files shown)


mariospr:
LGTM, with the disclaimer (that @pwithnall probably share) that this is kind of a terrible thing to do that we only assume because the very experimental nature of the project, which will only be based on wall clock time for now (thus this logic will have to be removed/changed in the near future anyway).

PR #21, opened 2018-09-07
Use libgsystemservice (as a submodule)

wjt:
This isn't currently packaged, and is theoretically API-unstable so I think a submodule is more appropriate.

https://phabricator.endlessm.com/T22281


wjt:
I have actually tested this now and it (a) builds with `dpkg-buildpackage -b` (which is not tested by Jenkins and which previously failed); and (b) works.

A caveat: as it stands, this branch will fail to build a `.deb` because some spurious `.test` files are installed by libgsystemservice even with `-Dinstalled_tests=false`. I filed https://gitlab.gnome.org/pwithnall/libgsystemservice/merge_requests/10 with a fix. As and when that is merged I will update the submodule in this PR.

I learnt several things about meson.

pwithnall:
Should the ‘WIP’ still be in the PR title? I won’t merge it because of that, but will review now.

I’ve merged https://gitlab.gnome.org/pwithnall/libgsystemservice/merge_requests/10 upstream.

wjt:
> Should the ‘WIP’ still be in the PR title? I won’t merge it because of that, but will review now.

Yes, because I was waiting for gss MR 10 to be merged to update the submodule!

PR #24, opened 2018-09-20
Add support for additional, out-of-tree PAYG providers

wjt:
We need to be able to integrate with existing PAYG ecosystems, which come with their own flavours of unlock codes and their own non-libre verification libraries.

Outstanding:

* [x] Adjust packaging. In particular, we'll need a -dev package which installs the .pc file and a few headers. I think the shared library can just live in the main daemon package.
* [x] Fill in some docs which are currently of the form "..." or "document me".
* [x] Test this and an out-of-tree provider plugin in situ.
* [x] Update README
* [x] Maybe other things

https://phabricator.endlessm.com/T23606


libeos-payg/meson.build
```diff
@@ -31,15 +37,40 @@ libeos_payg_resources = gnome.compile_resources(
...
+  description: 'Pay-as-you-go support for Endless OS',
+  version: meson.project_version(),
+  libraries: libeos_payg,
+  subdirs: libeos_payg_api_name,
+  variables: [
+    'pluginsdir=' + pluginsdir,
+  ],
+  # According to http://mesonbuild.com/Pkgconfig-module.html#pkggenerate,
+  # libeos_payg's dependencies will automatically be added to Requires.private
+  # with Meson >= 0.45, but this doesn't seem true in my testing.
```
pwithnall: A Meson bug? Filed a bug report?

libeos-payg/multi-task.c
```diff
@@ -92,6 +92,39 @@ epg_multi_task_return_boolean (GTask   *task,
...
+ * @task: (transfer none): the task
+ * @result: (transfer full): the result
+ * @result_destroy: a %GDestroyNotify function for @result
+ *
+ * Decrements the number of pending operations for @task by 1, causing it to
+ * return @result if no further operations are pending and no error has been
+ * encountered. If an error has been encountered, or further operations are
+ * pending, @result is immediately destroyed with @result_destroy.
+ *
+ * One way to use this function is to use a %GPtrArray as @task's task_data.
```
pwithnall: Nitpick: s/`%`/`#`/ for types. `%` should be used for enum members.

libeos-payg/multi-task.c
```diff
@@ -92,6 +92,39 @@ epg_multi_task_return_boolean (GTask   *task,
...
+ * Each time a subtask completes, add its output to the array, and call
+ * `epg_multi_task_return_pointer (task, g_ptr_array_ref (array),
+ * g_ptr_array_unref);`.
```
pwithnall: Nitpick: Missing `Since: …` line

libeos-payg/multi-task.h
```diff
@@ -30,6 +30,9 @@ void epg_multi_task_attach         (GTask         *task,
 void epg_multi_task_increment      (GTask         *task);
 void epg_multi_task_return_boolean (GTask         *task,
                                     gboolean       result);
+void epg_multi_task_return_pointer (GTask         *task,
```
pwithnall: Note: As discussed in person in London, it would be good if `EpgMultiTask` could be submitted upstream to GLib. However, that’s definitely not a prerequisite for this PR. I’m not going to do API review on it here, since the API seems appropriate for its use in this PR.

Upstream-wise, I think the API we discussed was to have the `MultiTask` expose all the return values from all its sub-tasks, and allow the caller to implement their own result accumulation; rather than trying to provide some generic mechanism for result accumulation.

libeos-payg/provider-loader.c
```diff
@@ -0,0 +1,310 @@
...
+                                              guint         property_id,
+                                              const GValue *value,
+                                              GParamSpec   *pspec);
+static void epg_provider_loader_constructed (GObject *object);
+static void epg_provider_loader_dispose (GObject *object);
+
+/**
+ * EpgProviderLoader:
+ *
+ * TODO: write me
```
pwithnall: Note: A wild `TODO` comment! :lion:

libeos-payg/provider-loader.c
```diff
@@ -0,0 +1,310 @@
...
+epg_provider_loader_constructed (GObject *object)
+{
+  EpgProviderLoader *self = EPG_PROVIDER_LOADER (object);
+
+  G_OBJECT_CLASS (epg_provider_loader_parent_class)->constructed (object);
+
+  if (self->module_dir == NULL)
+    self->module_dir = g_getenv ("EPG_MODULE_DIR");
+
+  if (self->module_dir == NULL)
```
pwithnall: Might want to squash `EPG_MODULE_DIR=""` down to `NULL` here and ignore it?

libeos-payg/provider-loader.c
```diff
@@ -0,0 +1,310 @@
...
+                              GAsyncResult *result,
+                              gpointer      user_data);
+static void maybe_return (GTask       *task,
+                          EpgProvider *provider);
+
+/**
+ * epg_provider_loader_load_async:
+ * @self: a %EpgProviderLoader
+ * @enabled: ...
+ * ...
```
pwithnall: A wild `...` comment! :tiger:

libeos-payg/provider-loader.c
```diff
@@ -0,0 +1,310 @@
...
+
+  if (!peas_engine_load_plugin (self->engine, plugin_info))
+    {
+      g_autoptr(GError) local_error = NULL;
+      gboolean ret = peas_plugin_info_is_available (plugin_info,
+                                                    &local_error);
+      if (ret)
+        {
+          /* Should not happen: peas_engine_load_plugin() should cause the
+           * PeasPluginInfo's error to be set.
```
pwithnall: If this ‘should not happen’, perhaps it should be an assertion?

libeos-payg/provider-loader.c
```diff
@@ -0,0 +1,310 @@
...
+                                       EPG_TYPE_PROVIDER))
+    return;
+
+  extension = peas_engine_create_extension (self->engine,
+                                            plugin_info,
+                                            EPG_TYPE_PROVIDER,
+                                            NULL);
+  if (extension == NULL)
+    {
+      /* Should not happen: we just checked that plugin_info can do this. */
```
pwithnall: Again, should this be an assertion?

libeos-payg/provider-loader.c
```diff
@@ -0,0 +1,310 @@
...
+}
+
+static void
+provider_init_cb (GObject      *source_object,
+                  GAsyncResult *result,
+                  gpointer      user_data)
+{
+  g_autoptr(EpgProvider) provider = EPG_PROVIDER (source_object);
+  g_autoptr(GTask) task = G_TASK (user_data);
+  g_autoptr(GError) error = NULL;
```
pwithnall: Nitpick: I prefer the name `local_error` to be used for local `GError*` variables, to distinguish it from `GError **error` arguments.

libeos-payg/provider-loader.c
```diff
@@ -0,0 +1,310 @@
...
+   * error should we return?
+   */
+  epg_multi_task_return_pointer (task,
+                                 g_ptr_array_ref (providers),
+                                 (GDestroyNotify) g_ptr_array_unref);
+}
+
+/**
+ * epg_provider_loader_load_finish:
+ *
```
pwithnall: Needs some documentation.

libeos-payg/provider-loader.c
```diff
@@ -0,0 +1,310 @@
...
+{
+  GPtrArray *providers = g_task_get_task_data (task);
+
+  if (provider != NULL)
+    g_ptr_array_add (providers, g_steal_pointer (&provider));
+
+  /* At present, we always return success, even if we found >= 1 plugin but all
+   * failed to load/initialize for some reason. Should we return an error in
+   * that case? If so: if two plugins are found and both fail to load, which
+   * error should we return?
```
pwithnall: I guess it depends where you want the policy of “exactly one provider must be loaded” to be implemented. Given that the code to load the internal fallback provider is not in the `ProviderLoader`, that suggests to me that the `ProviderLoader` should just return the array of providers it to loaded (which may be empty), and let the caller figure out what to do next (which might include erroring).

That appears to be what you’ve done anyway. Maybe just turn this comment into an explanation of that.

meson.build
```diff
@@ -22,6 +22,8 @@ localedir = join_paths(prefix, get_option('localedir'))
 includedir = join_paths(prefix, get_option('includedir'))
 localstatedir = join_paths(prefix, get_option('localstatedir'))
 sysconfdir = join_paths(prefix, get_option('sysconfdir'))
+pkglibdir = join_paths(libdir, meson.project_name()) # TODO: -1 suffix? soname?
```
pwithnall: Any suffix would be for the purposes of parallel-installability, and hence would have to include the major version of the project, and the major version of the plugin API. Those versions might be linked so that only one version number needs to be included.

I would recommend adding in a major version number suffix here, yes.

libeos-payg/service.c
```diff
@@ -125,22 +126,90 @@ epg_service_get_main_option_entries (GssService *service)
...
+  if (providers == NULL)
+    {
+      g_warning ("%s: Failed to load external providers: %s",
+                 G_STRFUNC, local_error->message);
+      g_clear_error (&local_error);
+    }
+  else
+    {
+      EpgProvider *provider = NULL;
+      guint i;
```
pwithnall: Nitpick: `gsize` for array iteration please

libeos-payg/service.c
```diff
@@ -125,22 +126,90 @@ epg_service_get_main_option_entries (GssService *service)
...
+                       G_STRFUNC, G_OBJECT_TYPE_NAME (provider));
+              provider = NULL;
+            }
+        }
+
+      /* Shut down all providers except the first enabled one (if any). */
+      for (i = 0; i < providers->len; i++)
+        {
+          EpgProvider *spare = EPG_PROVIDER (g_ptr_array_index (providers, i));
+          epg_provider_shutdown_async (spare, NULL, NULL, NULL);
```
pwithnall: I’m a little wary of this hanging async operation. There’s nothing which guarantees it’s completed before the `EpgService` is destroyed. Although unlikely, historically this kind of thing has been a source of bugs.

libeos-payg/service.c
```diff
@@ -184,15 +253,35 @@ manager_new_cb (GObject      *source_object,
...
 
+  epg_service_startup_complete (self, task, provider);
+}
+
+/**
+ * epg_service_startup_complete:
+ * @self:
+ * @task: (transfer none):
+ * @provider: (transfer full):
+ *
```
pwithnall: Needs documentation.

libeos-payg/service.c
```diff
@@ -184,15 +253,35 @@ manager_new_cb (GObject      *source_object,
...
+ */
+static void
+epg_service_startup_complete (EpgService  *self,
+                              GTask       *task,
+                              EpgProvider *provider)
+{
+  g_autoptr(GError) local_error = NULL;
+
+  self->provider = provider;
```
pwithnall: Nitpick: Might make things a little less likely to be broken in future to add a comment like `/* Take ownership */` above this assignment.

libeos-payg/service.c
```diff
@@ -184,15 +253,35 @@ manager_new_cb (GObject      *source_object,
...
-  self->provider = epg_manager_new_finish (result, &local_error);
-  if (self->provider == NULL)
+  provider = epg_manager_new_finish (result, &local_error);
+  if (provider == NULL)
     {
       g_task_return_error (task, g_steal_pointer (&local_error));
       return;
     }
 
+  epg_service_startup_complete (self, task, provider);
```
pwithnall: I’d do `g_steal_pointer (&provider)` instead of just passing `provider` in order to make it explicit that this takes ownership of it. Otherwise this could look like a memory leak.

libeos-payg/service.c
```diff
@@ -172,8 +241,8 @@ epg_service_startup_async (GssService          *service,
       return;
     }
 
-  epg_manager_new (enabled, NULL, NULL,
-                   cancellable,
+  GCancellable *cancellable = g_task_get_cancellable (task);
```
pwithnall: Maybe worth adding a comment here like `/* No provider plugin could be initialised. Fall back to #EpgManager. */`?

libeos-payg/service.c
```diff
@@ -125,22 +126,90 @@ epg_service_get_main_option_entries (GssService *service)
...
+      /* Shut down all providers except the first enabled one (if any). */
+      for (i = 0; i < providers->len; i++)
+        {
+          EpgProvider *spare = EPG_PROVIDER (g_ptr_array_index (providers, i));
+          epg_provider_shutdown_async (spare, NULL, NULL, NULL);
+        }
+
+      if (provider != NULL)
+        {
+          epg_service_startup_complete (self, task, provider);
```
pwithnall: I’d do `g_steal_pointer (&provider)` instead of just passing `provider` in order to make it explicit that this takes ownership of it. Otherwise this could look like a memory leak.

libeos-payg/service.c
pwithnall: Or maybe a bit further up by the `g_debug()` that kicks this all off.

libeos-payg/tests/manager.c
```diff
@@ -0,0 +1,348 @@
...
+{
+  g_clear_pointer (&fixture->expiry_time_path, remove_and_free_path);
+  g_clear_pointer (&fixture->used_codes_path, remove_and_free_path);
+
+  if (fixture->tmp_path != NULL)
+    {
+      /* The manager internally performs asynchronous saves in a GTask thread.
+       * There's currently no way to wait for them to finish. If it happens that
+       * one of the files gets recreated after being removed above but before
+       * we get here, removing the directory will fail. Ignore the error.
```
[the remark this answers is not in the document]
wjt: One of the later patches does add such a shutdown API, but I didn't arrange for the EpgManager implementation of it to wait for any other save operations to complete – maybe I should.

wjt:
test this please

wjt:
test this please

libeos-payg/tests/manager.c
```diff
@@ -0,0 +1,348 @@
...
+       gconstpointer data)
+{
+  g_autoptr(GError) error = NULL;
+
+  fixture->tmp_path = g_dir_make_tmp ("libeos-payg-tests-manager-XXXXXX", &error);
+  g_assert_no_error (error);
+  g_assert_nonnull (fixture->tmp_path);
+
+  fixture->tmp_dir = g_file_new_for_path (fixture->tmp_path);
+  g_assert_nonnull (fixture->tmp_dir);
```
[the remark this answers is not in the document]
wjt: Removed.

libeos-payg/tests/manager.c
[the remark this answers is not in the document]
wjt: Addressed by finding a GLib bug, working around it, and fixing this problem

libeos-payg/service.c
wjt: I think it is correct to match the type of `GPtrArray.len`:

```c
struct GPtrArray {
  gpointer *pdata;
  guint	    len;
};
```

libeos-payg/tests/manager.c
```diff
@@ -176,6 +184,112 @@ test_manager_load_empty (Fixture *fixture,
...
+  manager = manager_new_failable (fixture, &error);
+  g_assert_error (error, G_IO_ERROR, G_IO_ERROR_IS_DIRECTORY);
+  if (g_strstr_len (error->message, -1, path_p) == NULL)
+    g_error ("Error message '%s' does not contain state file path '%s'",
+             error->message, path_p);
+  g_assert_null (manager);
+  g_clear_error (&error);
+
+  /* Ideally, we would attempt to clean away the offending illegible 'file',
+   * but that is not currently the case.
```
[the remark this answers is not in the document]
wjt: It's just not implemented.

libeos-payg/tests/manager.c
```diff
@@ -0,0 +1,348 @@
...
+
+  while (result == NULL)
+    g_main_context_iteration (NULL, TRUE);
+
+  ret = epg_manager_save_state_finish (manager, result, &error);
+  g_assert_no_error (error);
+  g_assert_true (ret);
+}
+
+static void
```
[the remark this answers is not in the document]
wjt: I wrote some comments.

libeos-payg/manager.c
```diff
@@ -1070,28 +1050,21 @@ file_load_delete_cb (GObject      *source_object,
 {
   GFile *file = G_FILE (source_object);
   g_autoptr(GTask) task = G_TASK (user_data);
-  g_autoptr(GError) local_error = NULL;
-
-  /* Decrement the pending operation count. */
-  guint operation_count = GPOINTER_TO_UINT (g_task_get_task_data (task));
-  g_task_set_task_data (task, GUINT_TO_POINTER (--operation_count), NULL);
+  g_autoptr(GError) error = NULL;
```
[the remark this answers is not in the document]
wjt: Pushed a fixup to undo this.

libeos-payg/provider.c
```diff
@@ -0,0 +1,260 @@
...
+ * Returns: the UNIX timestamp when the current pay as you go top up will
+ *    expire, in seconds since the UNIX epoch
+ * Since: 0.2.0
+ */
+guint64
+epg_provider_get_expiry_time (EpgProvider *self)
+{
+  g_return_val_if_fail (EPG_IS_PROVIDER (self), 0);
+
+  return EPG_PROVIDER_GET_IFACE (self)->get_expiry_time (self);
```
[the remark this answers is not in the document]
wjt: I don't know which interface init function you mean: the only one needed by G_DEFINE_INTERFACE is epg_provider_default_init() and I think that's called on the interface itself, not for implementations of it. I added assertions to each wrapper function instead.

wjt:
bot, build

wjt:
test this please

libeos-payg/provider.c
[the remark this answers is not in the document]
pwithnall: I was misremembering `g_type_add_interface_check()`, but it should do what you want. Asserting in the wrapper functions works almost as well though.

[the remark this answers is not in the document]
wjt: This function is … not widely used! https://github.com/search?q=org%3AGNOME+g_type_add_interface_check&type=Code

libeos-payg/manager.c
```diff
@@ -1064,7 +1072,7 @@ static void file_save_delete_cb (GObject      *source_object,
                                  GAsyncResult *result,
                                  gpointer      user_data);
 
-void
+static void
```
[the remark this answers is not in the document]
wjt: Yes, they should have been part of `Extract an EpgProvider interface from EpgManager`. I can only realistically fix this by going back and amending the commits manually at merge time.

libeos-payg/multi-task.c
```diff
@@ -39,12 +39,27 @@ epg_multi_task_set (GTask *task,
...
+ * epg_multi_task_attach:
+ * @task: (transfer none): the task
+ * @pending: initial number of pending operations for @task.
+ *
+ * Initializes @task as having @pending operations pending.  The number of
+ * pending operations can be incremented with epg_multi_task_increment(), and
+ * decremented with epg_multi_task_return_boolean(),
+ * epg_multi_task_return_pointer(), and epg_multi_task_return_error().
+ *
+ * The number of pending operations is stored by occult means, leaving the
```
pwithnall: More documentation should be like this. ⛧

libeos-payg/tests/manager.c
```diff
@@ -408,6 +408,62 @@ test_manager_add_save_reload (Fixture *fixture,
   g_assert_cmpuint (now + 5, ==, expiry_after_reload);
 }
 
+/* test_manager_add_infinite_code:
+ *
+ * Tests that applying an infinite code sets the expiry time to infinitely far
+ * in the future, and that this survives both a restart and adding a
+ * non-infinite code afterwards.
```
pwithnall: Nice.

meson.build
```diff
@@ -79,7 +79,6 @@ test_args = [
   '-Wmissing-prototypes',
   '-Wnested-externs',
   '-Wno-error=cpp',
-  '-Wno-discarded-qualifiers',
```
pwithnall: Maybe explicitly switch this to `-Wdiscarded-qualifiers` to make it clear that it’s a warning we care about?

libeos-payg/manager.c
```diff
@@ -796,8 +814,15 @@ epg_manager_add_code (EpgProvider   *provider,
...
   clear_rate_limiting (self);
 
-  /* Kick off an asynchronous save. */
-  epg_manager_save_state_async (provider, self->cancellable, NULL, NULL);
+  /* Kick off an asynchronous save.
+   *
+   * FIXME: pass self->cancellable; see comment in
+   * epg_manager_shutdown_async().
+   */
+  g_assert_cmpuint (self->pending_internal_save_state_calls, <, G_MAXUINT64);
```
pwithnall: Use `g_assert()`, rather than `g_assert_*()`, in non-test code. `g_assert_*()` log through the test logging code, and are affected by various test settings. `g_assert()` is disableable by compiling with `G_DISABLE_ASSERT`, which people typically want for higher performance production code. It also isn’t affected by any of the test settings.

libeos-payg/manager.c
```diff
@@ -822,12 +850,45 @@ epg_manager_clear_code (EpgProvider  *provider,
...
   clear_expiry_timer (self);
 
-  /* Kick off an asynchronous save. */
-  epg_manager_save_state_async (provider, self->cancellable, NULL, NULL);
+  /* Kick off an asynchronous save.
+   *
+   * FIXME: pass self->cancellable; see comment in
+   * epg_manager_shutdown_async().
+   */
+  g_assert_cmpuint (self->pending_internal_save_state_calls, <, G_MAXUINT64);
```
pwithnall: Same here: use `g_assert()`.

libeos-payg/manager.c
```diff
@@ -822,12 +850,45 @@ epg_manager_clear_code (EpgProvider  *provider,
...
+  EpgProvider *provider = EPG_PROVIDER (self);
+  g_autoptr(GError) local_error = NULL;
+
+  if (!epg_manager_save_state_finish (provider, result, &local_error))
+    g_warning ("save_state failed: %s", local_error->message);
+
+  g_return_if_fail (self->pending_internal_save_state_calls > 0);
+  self->pending_internal_save_state_calls--;
+
+  if (self->pending_internal_save_state_calls == 0 &&
```
pwithnall: This block deserves a bit of a comment, I think, since it’s effectively kicking off another save operation directly after a previous one has (potentially successfully) finished.

libeos-payg/multi-task.c
wjt: Credit where credit's due: I stole this phrase from a [New Yorker review of Black Mirror](https://www.newyorker.com/magazine/2015/01/05/button-pusher) on the occasion of its legal release in the US:

> For a full day after watching the first episode (which I obtained through occult means, before Netﬂix made the show available to U.S. viewers), I felt disoriented, dropped on a new planet.

libeos-payg/manager.c
wjt: Hmm. Actually, this is only needed because if there was more than one pending call, it's not guaranteed that the later one "wins". Perhaps I should change all this logic to allow:

* at most 1 save operation in flight (with 1 or more tasks chained to it)
* a possibly-empty queue of requests to be serviced (ideally all at once) when the in-flight request completes

It seems like a lot of :keyboard: but it would have the advantage of being more robust in the face of unexpected power loss. I only have limited appetite for caring about this, though, because in ordinary use it's not possible to have >1 save at a time (since the UI only allows you to enter a single code, only when the machine is locked).

wjt: The less laborious option is to make shutdown_async something like:

```
if no saves in flight:
    return True # nothing to do
else:
    wait for last save to finish, propagate its result
```

This changes the behaviour in the case where writing back the state would fail, but the state hasn't changed so there's no need to write it back. It also ignores the problem that if two saves are in flight it's not defined which would finish first. But it is a lot easier…

wjt: I implemented:

```
if no saves in flight:
   save once more
else:
   wait for last to finish & propagate result
```

libeos-payg/service.c
wjt: Does this internal function need any more than this? I only wrote a docstring at all to have a place to put the transfer annotations. I am happy to just delete it.

libeos-payg/service.c
wjt: So what would you have this do – wait for all shutdowns to finish before pressing on with the main event?

libeos-payg/manager.c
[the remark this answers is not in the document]
pwithnall: I’m sorry to have brought it up.

libeos-payg/service.c
pwithnall: (If I could go back in time and change `GPtrArray.len` to be a `gsize`, I would. Then we could legitimately address every byte of address space in a `GPtrArray` and the world would somehow be a better place.)

libeos-payg/provider.c
```diff
@@ -90,20 +90,25 @@ epg_provider_default_init (EpgProviderInterface *iface)
...
+   * A regular expression which matches the format of codes expected by
+   * epg_provider_add_code() on this provider, in the dialect understood by
+   * GRegex. The expression should be anchored at each end with `^` and `$`.
    * Some examples:
    *
-   * - `"00000000"`: codes are eight arbitrary digits
-   * - `"*00000000000000#"`: codes begin with `*`, followed by 14 digits,
-   *   followed by `#`
+   * - `^\d{8}$`: codes are 8 arbitrary digits, such as `"12345678"` or
+   *   `"१२३४५६७८"` (use `[0-9]` if you mean ASCII digits)
```
pwithnall: I didn’t think it was possible to get this obscure. You win points.

libeos-payg/tests/manager.c
```diff
@@ -617,6 +705,30 @@ main (int    argc,
...
+
+  const char *partial_matches[] = { "1", "12", "1234567", NULL };
+  add_many ("/manager/code-format/partial",
+            test_manager_code_format_matches_partial,
+            partial_matches);
+
+  const char *non_matches[] = {
+      "\n", "a", "1a", "123\n", "a123",
+      /* '6' in various non-ASCII forms */
+      "६", "೬", "𝟨", "六",
```
pwithnall: I can see you’ve had a lot of fun here. This review has been a rollercoaster of emotion.

libeos-payg/tests/manager.c
```diff
@@ -595,6 +665,24 @@ test_manager_error_rate_limit (Fixture *fixture,
   g_assert_cmpuint (now + 5, ==, epg_provider_get_expiry_time (fixture->provider));
 }
 
+static void
+add_many (const char  *path_base,
+          void       (*func) (Fixture *, gconstpointer),
+          const char **data)
```
pwithnall: Are you a practizer of the occult art of functional programming, perchance?

libeos-payg/tests/plugins/meson.build
```diff
@@ -0,0 +1,42 @@
...
+      libtest_provider_dep,
+    ],
+  )
+
+  plugin_conf = configuration_data()
+  plugin_conf.set('suffix', suffix)
+  configure_file(
+    input: 'epg-test-provider.plugin.in',
+    output: 'epg-test-provider-' + suffix + '.plugin',
+    configuration: plugin_conf
```
pwithnall: Nitpick: I’d add a trailing comma here for future-proofness.

libeos-payg/tests/plugins/test-provider.c
```diff
@@ -0,0 +1,319 @@
...
+  gboolean enabled;
+} EpgTestProviderPrivate;
+
+typedef enum
+{
+  PROP_EXPIRY_TIME = 1,
+  PROP_ENABLED,
+  PROP_RATE_LIMIT_END_TIME,
+  PROP_CODE_FORMAT,
+} EpgProviderAngazaProperty;
```
pwithnall: I suspect this should be `EpgTestProviderProperty`.

libeos-payg/tests/plugins/test-provider.c
```diff
@@ -0,0 +1,319 @@
...
+
+  g_object_class_override_property (object_class, PROP_EXPIRY_TIME, "expiry-time");
+  g_object_class_override_property (object_class, PROP_ENABLED, "enabled");
+  g_object_class_override_property (object_class, PROP_RATE_LIMIT_END_TIME, "rate-limit-end-time");
+  g_object_class_override_property (object_class, PROP_CODE_FORMAT, "code-format");
+}
+
+static void
+epg_test_provider_async_initable_iface_init (gpointer g_iface,
+                                               gpointer iface_data)
```
pwithnall: Nitpick: Indentation problem (and in a few other places in this file).

libeos-payg/tests/plugins/test-provider.c
wjt: I think you're right!

libeos-payg/service.c
wjt: And we could address more than 80 GiB of instances of EpgProvider implementations :-)

wjt:
Rebased and squashed. Many conflicts but there was no diff to the pre-squashed version once complete, and `ninja test` passed after every conflict resolution.

PR #26, opened 2018-10-17
Add eos-payg-ctl

wjt:
This minimal tool is marginally easier to use than D-Feet to drive eos-paygd by hand and inspect its state.

https://phabricator.endlesssm.com/T24085


debian/control
```diff
@@ -94,6 +94,14 @@ Package: eos-payg-data
 Section: misc
 Architecture: all
 Depends:
+ eos-paygd (= ${binary:Version}),
```
pwithnall: I would consider making this a separate `eos-payg-tools` package, but I don’t mind either way.

eos-payg-ctl/eos-payg-ctl
```diff
@@ -0,0 +1,112 @@
+#!/usr/bin/python3
```
pwithnall: Could do with a license and encoding header.
```suggestion
#!/usr/bin/python3
# -*- coding: utf-8 -*-
#
# Copyright © 2018 Endless Mobile, Inc.
#
# This Source Code Form is subject to the terms of the Mozilla Public License,
# v. 2.0. If a copy of the MPL was not distributed with this file, You can
# obtain one at https://mozilla.org/MPL/2.0/.
#
# Alternatively, the contents of this file may be used under the terms of the
# GNU Lesser General Public License Version 2.1 or later (the "LGPL"), in
# which case the provisions of the LGPL are applicable instead of those above.
# If you wish to allow use of your version of this file only under the terms
# of the LGPL, and not to allow others to use your version of this file under
# the terms of the MPL, indicate your decision by deleting the provisions
# above and replace them with the notice and other provisions required by the
# LGPL. If you do not delete the provisions above, a recipient may use your
# version of this file under the terms of either the MPL or the LGPL.

```

eos-payg-ctl/eos-payg-ctl
```diff
@@ -0,0 +1,112 @@
...
+            formatted = formatter(value)
+        print('{:>{}}: {}'.format(key, width, formatted))
+
+
+@contextlib.contextmanager
+def __exit_on_payg_error():
+    try:
+        yield
+    except GLib.GError as e:
+        # The Right Thing To Do would be to make libeos-payg introspectable, so
```
pwithnall:
```suggestion
        # FIXME: The Right Thing To Do would be to make libeos-payg introspectable, so
```

eos-payg-ctl/eos-payg-ctl
```diff
@@ -0,0 +1,112 @@
...
+        Gio.DBusProxyFlags.NONE,
+        None,
+        BUS_NAME,
+        OBJECT_PATH,
+        INTERFACE,
+        None,  # cancellable
+    )
+
+
+def command_status(proxy):
```
pwithnall: This function is a masterpiece of Pythonism :ok_hand:

eos-payg-ctl/eos-payg-ctl
```diff
@@ -0,0 +1,112 @@
...
+    parser = argparse.ArgumentParser()
+    parser.set_defaults(function=command_status)
+    subparsers = parser.add_subparsers()
+
+    status = subparsers.add_parser('status')
+    status.set_defaults(function=command_status)
+
+    add_code = subparsers.add_parser('add-code')
+    add_code.set_defaults(function=command_add_code)
+    add_code.add_argument('code')
```
pwithnall: Could do with a description. The rest of the arguments could do with some descriptions and help too.

eos-payg-ctl/meson.build
```diff
@@ -0,0 +1,6 @@
+configure_file(
+  input: 'eos-payg-ctl',
+  output: 'eos-payg-ctl',
+  copy: true,
+  install_dir: bindir,
+)
```
pwithnall: Could this just be `install_data` instead?
```
install_data('eos-payg-ctl',
  install_dir: bindir,
  install_mode: 'rwxr-xr-x',
)
```

pwithnall:
> This minimal tool is marginally easier to use than D-Feet to drive eos-paygd by hand and inspect its state.
> 
> https://phabricator.endlesssm.com/T24085

Typo in the Phabricator link.

eos-payg-ctl/eos-payg-ctl
wjt: FWIW UTF-8 is the default for Python 3. https://docs.python.org/3/howto/unicode.html#unicode-literals-in-python-source-code

debian/control
wjt: I did consider this but I decided that adding 1 more package and then go list it in eos-meta isn't a net win.

Agreed that a manpage would be nice but it's a marginal win over improving the `--help` output (at which point one might generate it at package build time with help2man).

eos-payg-ctl/eos-payg-ctl
```diff
@@ -0,0 +1,140 @@
...
+        'RateLimitEndTime': __format_utc_timestamp,
+    }
+
+    for key, value in sorted(props.items()):
+        try:
+            formatter = formatters[key]
+        except KeyError:
+            formatted = value
+        else:
+            formatted = formatter(value)
```
wjt: I almost wrote this as

```python
formatted = formatters.get(key, lambda x: x)(value)
```

and actually I think that's clearer now I type it out…

wjt:
I kept trying to run this tool as `eos-updater-ctl` while developing it and wondering why tab-completion didn't work, too…

PR #28, opened 2018-11-20
T24300 Use CLOCK_BOOTTIME during runtime

mwleeds:

### DIFF
--- a/libeos-payg/boottime-source.c
+++ b/libeos-payg/boottime-source.c
@@ -1,0 +1,178 @@
+/* -*- mode: C; c-file-style: "gnu"; indent-tabs-mode: nil; -*-
+ *
+ * Copyright © 2018 Endless Mobile, Inc.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at https://mozilla.org/MPL/2.0/.
+ *
+ * Alternatively, the contents of this file may be used under the terms of the
+ * GNU Lesser General Public License Version 2.1 or later (the "LGPL"), in
+ * which case the provisions of the LGPL are applicable instead of those above.
+ * If you wish to allow use of your version of this file only under the terms
+ * of the LGPL, and not to allow others to use your version of this file under
+ * the terms of the MPL, indicate your decision by deleting the provisions
+ * above and replace them with the notice and other provisions required by the
+ * LGPL. If you do not delete the provisions above, a recipient may use your
+ * version of this file under the terms of either the MPL or the LGPL.
+ */
+
+#include "config.h"
+
+#include <errno.h>
+#include <inttypes.h>
+#include <sys/timerfd.h>
+#include <unistd.h>
+#include <glib.h>
+#include <glib/gstdio.h>
+#include <gio/gio.h>
+#include <libeos-payg/boottime-source.h>
+
+typedef struct {
+  GSource parent;
+
+  int fd;
+  gpointer tag;
+} EpgBoottimeSource;
+
+static gboolean
+epg_boottime_source_check (GSource *source)
+{
+  EpgBoottimeSource *self = (EpgBoottimeSource *) source;
+
+  return g_source_query_unix_fd (source, self->tag) != 0;
+}
+
+static gboolean
+epg_boottime_source_dispatch (GSource    *source,
+                              GSourceFunc callback,
+                              gpointer    user_data)
+{
+  EpgBoottimeSource *self = (EpgBoottimeSource *) source;
+  uint64_t n_expirations = 0;
+
+  if (callback == NULL)
+    {
+      g_warning ("Boottime source dispatched without callback. "
+                 "You must call g_source_set_callback().");
+      return G_SOURCE_REMOVE;
+    }
+
+  /* Must read from the FD to reset its ready state. */
+  if (read (self->fd, &n_expirations, sizeof (n_expirations)) < 0)
+    {
+      g_warning ("read() failed for timerfd: %s",
+                 g_strerror (errno));
+      return G_SOURCE_REMOVE;
+    }
+
+  return callback (user_data);
+}
+
+static void
+epg_boottime_source_finalize (GSource *source)
+{
+  EpgBoottimeSource *self = (EpgBoottimeSource *) source;
+  g_autoptr(GError) local_error = NULL;
+
+  if (!g_close (self->fd, &local_error))
+    g_warning ("Failed to close timerfd: %s", local_error->message);
+  self->fd = -1;
+}
+
+static const GSourceFuncs epg_boottime_source_funcs = {
+  .check = epg_boottime_source_check,
+  .dispatch = epg_boottime_source_dispatch,
+  .finalize = epg_boottime_source_finalize,
+};
+
+/**
+ * epg_boottime_source_new:
+ * @interval_ms: the timeout interval, in ms
+ * @error: return location for a #GError, or %NULL
+ *
+ * Like g_timeout_source_new(), but uses `CLOCK_BOOTTIME` to account for time
+ * when the system is suspended.
+ *
+ * If @interval_ms is set to zero, the #GSource will be ready the next time
+ * it's checked.
+ *
+ * @error will be set to a #GIOError if, for example, the process runs out
+ * of file descriptors.
+ *
+ * Returns: (transfer full): a new `CLOCK_BOOTTIME` #GSource, or %NULL with
+ *   @error set
+ * Since: 0.2.1
+ */
+GSource *
+epg_boottime_source_new (guint    interval_ms,
+                         GError **error)
+{
+  g_autoptr(GSource) source = NULL;
+  EpgBoottimeSource *self = NULL;
+  int fd;
+  struct itimerspec its = {
+    .it_interval = {
+      .tv_sec = interval_ms / 1000,
+      .tv_nsec = (interval_ms % 1000) * (1000 * 1000),
+    }
+  };
+
+  its.it_value = its.it_interval;
+
+  /* Set the GError if timerfd_create() fails because it could be e.g. ENFILE
+   * which we should handle gracefully */
+  fd = timerfd_create (CLOCK_BOOTTIME, TFD_CLOEXEC | TFD_NONBLOCK);
+  if (G_UNLIKELY (fd < 0))
+    {
+      g_set_error (error, G_IO_ERROR, g_io_error_from_errno (errno),
+                   "timerfd_create (CLOCK_BOOTTIME, TFD_CLOEXEC | TFD_NONBLOCK) failed: %s",
+                   g_strerror (errno));
+      return NULL;
+    }
+
+  /* But use g_error() if timerfd_settime() fails which likely would indicate
+   * programmer error */
+  /* FIXME: If this is upstreamed to GLib, handle EINVAL gracefully */
+  if (G_UNLIKELY (timerfd_settime (fd,
+                                   0,
+                                   &its,
+                                   NULL /* old_value */) < 0))
+    g_error ("timerfd_settime() failed: %s",
+             g_strerror (errno));
+
+  source = g_source_new ((GSourceFuncs *)&epg_boottime_source_funcs,
+                         sizeof (EpgBoottimeSource));
+  self = (EpgBoottimeSource *) source;
+  self->fd = fd;
+  self->tag = g_source_add_unix_fd (source, fd,
+                                    G_IO_IN | G_IO_HUP | G_IO_ERR | G_IO_NVAL);
+
+  return g_steal_pointer (&source);
+}
+
+/**
+ * epg_get_boottime:
+ *
+ * Like g_get_monotonic_time(), but also includes any time the system is
+ * suspended. Uses CLOCK_BOOTTIME, hence the name, but is not guaranteed to be
+ * the time since boot.
+ *
+ * Returns: the time since some unspecified starting point, in microseconds
+ * Since: 0.2.1
+ */
+gint64
+epg_get_boottime (void)
+{
+  struct timespec ts;
+  gint result;
+
+  result = clock_gettime (CLOCK_BOOTTIME, &ts);
+
+  if (G_UNLIKELY (result != 0))
+    g_error ("clock_gettime (CLOCK_BOOTTIME) failed: %s",
+             g_strerror (errno));
+
+  return (((gint64) ts.tv_sec) * 1000000) + (ts.tv_nsec / 1000);
+}
+
--- a/libeos-payg/clock.c
+++ b/libeos-payg/clock.c
@@ -1,0 +1,102 @@
+/* -*- mode: C; c-file-style: "gnu"; indent-tabs-mode: nil; -*-
+ *
+ * Copyright © 2018 Endless Mobile, Inc.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at https://mozilla.org/MPL/2.0/.
+ *
+ * Alternatively, the contents of this file may be used under the terms of the
+ * GNU Lesser General Public License Version 2.1 or later (the "LGPL"), in
+ * which case the provisions of the LGPL are applicable instead of those above.
+ * If you wish to allow use of your version of this file only under the terms
+ * of the LGPL, and not to allow others to use your version of this file under
+ * the terms of the MPL, indicate your decision by deleting the provisions
+ * above and replace them with the notice and other provisions required by the
+ * LGPL. If you do not delete the provisions above, a recipient may use your
+ * version of this file under the terms of either the MPL or the LGPL.
+ */
+
+#include <clock.h>
+#include <glib.h>
+
+G_DEFINE_INTERFACE (EpgClock, epg_clock, G_TYPE_OBJECT)
+
+static void
+epg_clock_default_init (EpgClockInterface *iface)
+{
+}
+
+/**
+ * epg_clock_get_wallclock_time:
+ * @self: an #EpgClock
+ *
+ * Like g_get_real_time(), this returns the wall clock time, but the
+ * implementation determines what type of clock to use and seconds are used
+ * instead of microseconds.
+ *
+ * Returns: the number of seconds since January 1, 1970 UTC
+ * Since: 0.2.1
+ */
+gint64
+epg_clock_get_wallclock_time (EpgClock *self)
+{
+  EpgClockInterface *iface;
+
+  g_return_val_if_fail (EPG_IS_CLOCK (self), -1);
+
+  iface = EPG_CLOCK_GET_IFACE (self);
+  g_assert (iface->get_wallclock_time != NULL);
+  return iface->get_wallclock_time (self);
+}
+
+/**
+ * epg_clock_get_time:
+ * @self: an #EpgClock
+ *
+ * This returns the time based on the same clock used for
+ * epg_clock_source_new_seconds()
+ *
+ * Returns: the time since some unspecified starting point, in seconds
+ * Since: 0.2.1
+ */
+gint64
+epg_clock_get_time (EpgClock *self)
+{
+  EpgClockInterface *iface;
+
+  g_return_val_if_fail (EPG_IS_CLOCK (self), -1);
+
+  iface = EPG_CLOCK_GET_IFACE (self);
+  g_assert (iface->get_time != NULL);
+  return iface->get_time (self);
+}
+
+/**
+ * epg_clock_source_new_seconds:
+ * @self: an #EpgClock
+ * @interval: the timeout interval in seconds
+ * @error: return location for a #GError, or %NULL
+ *
+ * Like g_timeout_source_new_seconds(), but the implementation determines what
+ * type of clock to use. If @interval is set to zero, the source will be ready
+ * the next time it's checked.
+ *
+ * Returns: (transfer full): the newly-created timeout source, or %NULL with
+ *   @error set
+ * Since: 0.2.1
+ */
+GSource *
+epg_clock_source_new_seconds (EpgClock  *self,
+                              guint      interval,
+                              GError   **error)
+{
+  EpgClockInterface *iface;
+
+  g_return_val_if_fail (EPG_IS_CLOCK (self), NULL);
+  g_return_val_if_fail (error == NULL || *error == NULL, NULL);
+
+  iface = EPG_CLOCK_GET_IFACE (self);
+  g_assert (iface->source_new_seconds != NULL);
+  return iface->source_new_seconds (self, interval, error);
+}
--- a/libeos-payg/fake-clock.c
+++ b/libeos-payg/fake-clock.c
@@ -1,0 +1,239 @@
+/* -*- mode: C; c-file-style: "gnu"; indent-tabs-mode: nil; -*-
+ *
+ * Copyright © 2018 Endless Mobile, Inc.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at https://mozilla.org/MPL/2.0/.
+ *
+ * Alternatively, the contents of this file may be used under the terms of the
+ * GNU Lesser General Public License Version 2.1 or later (the "LGPL"), in
+ * which case the provisions of the LGPL are applicable instead of those above.
+ * If you wish to allow use of your version of this file only under the terms
+ * of the LGPL, and not to allow others to use your version of this file under
+ * the terms of the MPL, indicate your decision by deleting the provisions
+ * above and replace them with the notice and other provisions required by the
+ * LGPL. If you do not delete the provisions above, a recipient may use your
+ * version of this file under the terms of either the MPL or the LGPL.
+ */
+
+#include <libeos-payg/fake-clock.h>
+#include <libeos-payg/boottime-source.h>
+
+#define MSEC_PER_SEC 1000
+
+struct _EpgFakeClock
+{
+  GObject parent_instance;
+
+  gint64 time_secs;
+  gint64 wallclock_time_secs;
+};
+
+static void clock_iface_init (EpgClockInterface *iface,
+                              gpointer           iface_data);
+
+G_DEFINE_TYPE_WITH_CODE (EpgFakeClock, epg_fake_clock, G_TYPE_OBJECT,
+                         G_IMPLEMENT_INTERFACE (EPG_TYPE_CLOCK, clock_iface_init))
+
+static gint64
+epg_fake_clock_get_wallclock_time (EpgClock *clock)
+{
+  EpgFakeClock *self = (EpgFakeClock *)clock;
+
+  g_return_val_if_fail (EPG_IS_FAKE_CLOCK (self), -1);
+
+  /* Initialize to a realistic value without introducing non-determinism */
+  if (self->wallclock_time_secs == 0)
+    self->wallclock_time_secs = 1231006505;
+
+  return self->wallclock_time_secs;
+}
+
+/**
+ * epg_fake_clock_set_wallclock_time:
+ * @self: an #EpgFakeClock
+ * @new_time: a positive value to set the clock to
+ *
+ * Set the time to @new_time so that value will be returned by
+ * epg_clock_get_wallclock_time() until the next call to this function. This is
+ * intended to be used by unit tests.
+ *
+ * Since: 0.2.1
+ */
+void
+epg_fake_clock_set_wallclock_time (EpgFakeClock *self,
+                                   gint64        new_time)
+{
+  g_return_if_fail (EPG_IS_FAKE_CLOCK (self));
+  g_return_if_fail (new_time > 0);
+
+  self->wallclock_time_secs = new_time;
+}
+
+static gint64
+epg_fake_clock_get_time (EpgClock *clock)
+{
+  EpgFakeClock *self = (EpgFakeClock *)clock;
+
+  g_return_val_if_fail (EPG_IS_FAKE_CLOCK (self), -1);
+
+  /* Initialize to a realistic value without introducing non-determinism */
+  if (self->time_secs == 0)
+    self->time_secs = 424242;
+
+  return self->time_secs;
+}
+
+/**
+ * epg_fake_clock_set_time:
+ * @self: an #EpgFakeClock
+ * @new_time: a positive value to set the clock to
+ *
+ * Set the time to @new_time so that value will be returned by
+ * epg_clock_get_time() until the next call to this function. This is intended
+ * to be used by unit tests.
+ *
+ * Since: 0.2.1
+ */
+void
+epg_fake_clock_set_time (EpgFakeClock *self,
+                         gint64       new_time)
+{
+  g_return_if_fail (EPG_IS_FAKE_CLOCK (self));
+  g_return_if_fail (new_time > 0);
+
+  self->time_secs = new_time;
+}
+
+typedef struct {
+  GSource parent;
+
+  EpgFakeClock *clock; /* owned */
+  gint64 ready_time_secs;
+} EpgFakeSource;
+
+static gboolean
+epg_fake_source_check (GSource *source)
+{
+  EpgFakeSource *self = (EpgFakeSource *) source;
+
+  return epg_clock_get_time (EPG_CLOCK (self->clock)) >= self->ready_time_secs;
+}
+
+static gboolean
+epg_fake_source_dispatch (GSource    *source,
+                          GSourceFunc callback,
+                          gpointer    user_data)
+{
+  EpgFakeSource *self = (EpgFakeSource *) source;
+
+  if (callback == NULL)
+    {
+      g_warning ("Fake source dispatched without callback. "
+                 "You must call g_source_set_callback().");
+      return G_SOURCE_REMOVE;
+    }
+
+  self->ready_time_secs = G_MAXINT64;
+
+  return callback (user_data);
+}
+
+static void
+epg_fake_source_finalize (GSource *source)
+{
+  EpgFakeSource *self = (EpgFakeSource *) source;
+
+  g_clear_object (&self->clock);
+}
+
+static const GSourceFuncs epg_fake_source_funcs = {
+  .check = epg_fake_source_check,
+  .dispatch = epg_fake_source_dispatch,
+  .finalize = epg_fake_source_finalize,
+};
+
+/**
+ * epg_fake_clock_source_new_seconds:
+ * @clock: an #EpgClock
+ * @interval: the timeout interval in seconds
+ * @error: a return location for #GError
+ *
+ * This creates a GSource which will be ready when @interval seconds have
+ * passed according to epg_fake_clock_get_time().
+ *
+ * Returns: (transfer full) (nullable): the newly created #GSource
+ */
+static GSource *
+epg_fake_clock_source_new_seconds (EpgClock  *clock,
+                                   guint      interval,
+                                   GError   **error)
+{
+  EpgFakeClock *self = (EpgFakeClock *)clock;
+  g_autoptr(GSource) source = NULL;
+  EpgFakeSource *fake_source = NULL;
+  gint64 current_time = epg_clock_get_time (EPG_CLOCK (self));
+
+  g_return_val_if_fail (EPG_IS_FAKE_CLOCK (self), NULL);
+  g_return_val_if_fail (interval <= G_MAXINT64 - current_time, NULL);
+
+  source = g_source_new ((GSourceFuncs *)&epg_fake_source_funcs,
+                         sizeof (EpgFakeSource));
+  fake_source = (EpgFakeSource *)source;
+  fake_source->clock = g_object_ref (self);
+  fake_source->ready_time_secs = current_time + interval;
+  return g_steal_pointer (&source);
+}
+
+static void
+clock_iface_init (EpgClockInterface *iface,
+                  gpointer           iface_data)
+{
+  iface->get_wallclock_time = epg_fake_clock_get_wallclock_time;
+  iface->get_time = epg_fake_clock_get_time;
+  iface->source_new_seconds = epg_fake_clock_source_new_seconds;
+}
+
+static void
+epg_fake_clock_class_init (EpgFakeClockClass *klass)
+{
+}
+
+static void
+epg_fake_clock_init (EpgFakeClock *self)
+{
+}
+
+/**
+ * epg_fake_clock_new:
+ * @optional_time: A positive time value, or -1
+ * @optional_wallclock_time: A positive wallclock time value, or -1
+ *
+ * EpgFakeClock is a clock that can be arbitrarily set forward or backward, and
+ * is intended to be used for unit tests. It initializes with the values of
+ * @optional_time and @optional_wallclock_time if set, and otherwise with
+ * realistic, static values for both. The two different time values are roughly
+ * analogous to time according to the `CLOCK_BOOTTIME` kernel clock, and time
+ * according to the user-visible system clock.
+ *
+ * Once these values are initialized, they stand still and only
+ * change when epg_fake_clock_set_time() or epg_fake_clock_set_wallclock_time()
+ * is used.
+ *
+ * Returns: (transfer full): the newly-created #EpgFakeClock
+ * Since: 0.2.1
+ */
+EpgFakeClock *
+epg_fake_clock_new (gint64 optional_time,
+                    gint64 optional_wallclock_time)
+{
+  EpgFakeClock *new_clock = g_object_new (EPG_TYPE_FAKE_CLOCK, NULL);
+
+  if (optional_time > 0)
+    epg_fake_clock_set_time (new_clock, optional_time);
+  if (optional_wallclock_time > 0)
+    epg_fake_clock_set_wallclock_time (new_clock, optional_wallclock_time);
+
+  return new_clock;
+}
--- a/libeos-payg/manager.c
+++ b/libeos-payg/manager.c
@@ -58,7 +58,6 @@ static gboolean epg_manager_init_finish (GAsyncInitable  *initable,
 
 static gboolean    epg_manager_add_code   (EpgProvider  *provider,
                                            const gchar  *code_str,
-                                           guint64       now_secs,
                                            GError      **error);
 static gboolean    epg_manager_clear_code (EpgProvider  *provider,
                                            GError      **error);
@@ -684,8 +683,7 @@ set_expiry_time (EpgManager *self,
  * would overflow, set the expiry time as high as possible. Everything is
  * handled in seconds.
  *
- * @now will typically be the current value of CLOCK_BOOTTIME.
- * It is parameterised to allow easy testing. */
+ * @now will typically be the current value of CLOCK_BOOTTIME. */
 static void
 extend_expiry_time (EpgManager *self,
                     guint64     now_secs,
@@ -799,11 +797,11 @@ used_codes_sort_cb (gconstpointer a,
 static gboolean
 epg_manager_add_code (EpgProvider   *provider,
                       const gchar  *code_str,
-                      guint64       now_secs,
                       GError      **error)
 {
   EpgManager *self = EPG_MANAGER (provider);
   g_autoptr(GError) local_error = NULL;
+  guint64 now_secs;
 
   g_return_val_if_fail (EPG_IS_MANAGER (provider), FALSE);
   g_return_val_if_fail (code_str != NULL, FALSE);
@@ -812,6 +810,7 @@ epg_manager_add_code (EpgProvider   *provider,
   if (!check_enabled (self, error))
     return FALSE;
 
+  now_secs = epg_clock_get_time (self->clock);
   if (!check_rate_limiting (self, now_secs, error))
     return FALSE;
 
--- a/libeos-payg/manager.c
+++ b/libeos-payg/manager.c
@@ -1269,6 +1269,17 @@ file_load_cb (GObject      *source_object,
           else
             set_expiry_time (self, now_secs, self->last_save_expiry_secs - unaccounted_time);
         }
+
+      /* Kick off an asynchronous save. Otherwise an unclean shutdown would
+       * cause us to consume credit for the same time period again.
+       *
+       * FIXME: pass self->cancellable; see comment in
+       * epg_manager_shutdown_async().
+       */
+      g_assert (self->pending_internal_save_state_calls < G_MAXUINT64);
+      self->pending_internal_save_state_calls++;
+      epg_manager_save_state_async (EPG_PROVIDER (self), NULL,
+                                    internal_save_state_cb, NULL);
     }
 
   epg_multi_task_return_boolean (task, TRUE);
--- a/libeos-payg/manager.c
+++ b/libeos-payg/manager.c
@@ -1015,6 +1015,24 @@ epg_manager_init_async (GAsyncInitable      *initable,
   epg_multi_task_return_boolean (task, TRUE);
 }
 
+/* Convert bytes into a guint64 assuming host endianness */
+/* @bytes must have length 8 */
+static guint64
+get_guint64_from_bytes (const guint8 *bytes)
+{
+  g_assert (bytes != NULL);
+
+  union
+    {
+      guint64 u64;
+      gchar u8[8];
+    } number_union;
+
+  memcpy (number_union.u8, bytes, sizeof (number_union.u8));
+
+  return number_union.u64;
+}
+
 static void
 file_load_cb (GObject      *source_object,
               GAsyncResult *result,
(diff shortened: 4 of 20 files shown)


libeos-payg/service.c
```diff
@@ -229,8 +231,9 @@ provider_get_first_enabled_cb (GObject      *source_object,
       return;
     }
 
+  clock = epg_real_clock_new ();
   GCancellable *cancellable = g_task_get_cancellable (task);
-  epg_manager_new (enabled, NULL, NULL, cancellable,
+  epg_manager_new (enabled, NULL, NULL, (EpgClock *)clock, cancellable,
```
pwithnall: Nitpick: Use `EPG_CLOCK (clock)` so you get the dynamic type check.

libeos-payg/boottime-source.c
```diff
@@ -0,0 +1,154 @@
...
+{
+  EpgBoottimeSource *self = (EpgBoottimeSource *) source;
+
+  return g_source_query_unix_fd (source, self->tag) != 0;
+}
+
+static gboolean
+epg_boottime_source_dispatch (GSource    *source,
+                                      GSourceFunc callback,
+                                      gpointer    user_data)
```
pwithnall: Nitpick: Indentation problem.

libeos-payg/boottime-source.c
```diff
@@ -0,0 +1,154 @@
...
+{
+  EpgBoottimeSource *self = (EpgBoottimeSource *) source;
+  uint64_t n_expirations = 0;
+
+  if (!callback)
```
pwithnall:
```suggestion
  if (callback == NULL)
```

libeos-payg/boottime-source.c
```diff
@@ -0,0 +1,154 @@
...
+    {
+      g_warning ("Boottime source dispatched without callback. "
+                 "You must call g_source_set_callback().");
+      return G_SOURCE_REMOVE;
+    }
+
+  /* Must read from the FD to reset its ready state. */
+  if (read (self->fd, &n_expirations, sizeof n_expirations) < 0)
```
pwithnall:
```suggestion
  if (read (self->fd, &n_expirations, sizeof (n_expirations)) < 0)
```

libeos-payg/boottime-source.c
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+  return callback (user_data);
+}
+
+static void
+epg_boottime_source_finalize (GSource *source)
+{
+  EpgBoottimeSource *self = (EpgBoottimeSource *) source;
+  g_autoptr(GError) error = NULL;
```
pwithnall:
```suggestion
  g_autoptr(GError) local_error = NULL;
```

libeos-payg/boottime-source.c
```diff
@@ -0,0 +1,154 @@
...
+
+  if (!g_close (self->fd, &error))
```
pwithnall:
```suggestion
  if (!g_close (self->fd, &local_error))
```

libeos-payg/boottime-source.c
```diff
@@ -0,0 +1,154 @@
...
+    g_warning ("failed to close timerfd: %s", error->message);
```
pwithnall:
```suggestion
    g_warning ("Failed to close timerfd: %s", local_error->message);
```

libeos-payg/boottime-source.c
```diff
@@ -0,0 +1,154 @@
...
+  self->fd = -1;
+}
+
+static GSourceFuncs epg_boottime_source_funcs = {
```
pwithnall:
```suggestion
static const GSourceFuncs epg_boottime_source_funcs = {
```

libeos-payg/boottime-source.c
```diff
@@ -0,0 +1,154 @@
...
+  .check = epg_boottime_source_check,
+  .dispatch = epg_boottime_source_dispatch,
+  .finalize = epg_boottime_source_finalize,
+};
+
+/**
+ * epg_boottime_source_new():
```
pwithnall:
```suggestion
 * epg_boottime_source_new:
 * @interval_ms: timeout interval in milliseconds
```

libeos-payg/boottime-source.c
```diff
@@ -0,0 +1,154 @@
...
+ *
+ * Like g_timeout_source_new(), but uses CLOCK_BOOTTIME to account for time
+ * when the system is suspended.
+ **/
```
pwithnall:
```suggestion
 *
 * Returns: (transfer full): a new `CLOCK_BOOTTIME` #GSource
 * Since: 0.2.1
 */
```

libeos-payg/boottime-source.c
```diff
@@ -0,0 +1,154 @@
...
+static GSourceFuncs epg_boottime_source_funcs = {
+  .check = epg_boottime_source_check,
+  .dispatch = epg_boottime_source_dispatch,
+  .finalize = epg_boottime_source_finalize,
+};
+
+/**
+ * epg_boottime_source_new():
+ *
+ * Like g_timeout_source_new(), but uses CLOCK_BOOTTIME to account for time
```
pwithnall:
```suggestion
 * Like g_timeout_source_new(), but uses `CLOCK_BOOTTIME` to account for time
```

libeos-payg/boottime-source.c
```diff
@@ -0,0 +1,154 @@
...
+    .it_interval = {
+      .tv_sec = interval_ms / 1000,
+      .tv_nsec = (interval_ms % 1000) * (1000 * 1000),
+    }
+  };
+
+  its.it_value = its.it_interval;
+
+  fd = timerfd_create (CLOCK_BOOTTIME, TFD_CLOEXEC);
+  if G_UNLIKELY (fd < 0)
```
pwithnall:
```suggestion
  if (G_UNLIKELY (fd < 0))
```

libeos-payg/boottime-source.c
```diff
@@ -0,0 +1,154 @@
...
+    g_error ("timerfd_create (CLOCK_BOOTTIME, TFD_CLOEXEC) failed: %s",
+             g_strerror (errno));
+
+  if G_UNLIKELY (timerfd_settime (fd,
```
pwithnall:
```suggestion
  if (G_UNLIKELY (timerfd_settime (fd,
```

libeos-payg/boottime-source.c
```diff
@@ -0,0 +1,154 @@
...
+                                  0 /*TFD_NONBLOCK */,
+                                  &its,
+                                  NULL /* old_value */) < 0)
```
pwithnall:
```suggestion
                                  NULL /* old_value */) < 0))
```

libeos-payg/boottime-source.c
```diff
@@ -0,0 +1,154 @@
...
+
+/**
+ * epg_get_boottime:
+ *
+ * Like g_get_monotonic_time(), but also includes any time the system is
+ * suspended. Uses CLOCK_BOOTTIME, hence the name, but is not guaranteed to be
+ * the time since boot.
+ *
+ * Returns: the time since some unspecified starting point, in microseconds
+ **/
```
pwithnall:
```suggestion
 * Since: 0.2.1
 */
```

libeos-payg/boottime-source.c
```diff
@@ -0,0 +1,154 @@
...
+gint64
+epg_get_boottime (void)
+{
+  struct timespec ts;
+  gint result;
+
+  result = clock_gettime (CLOCK_BOOTTIME, &ts);
+
+  if G_UNLIKELY (result != 0)
```
pwithnall:
```suggestion
  if (G_UNLIKELY (result != 0))
```

libeos-payg/clock.c
```diff
@@ -0,0 +1,90 @@
...
+
+/**
+ * epg_clock_get_wallclock_time:
+ * @self: an #EpgClock
+ *
+ * Like g_get_real_time(), this returns the wall clock time, but the
+ * implementation determines what type of clock to use.
+ *
+ * Returns: the number of microseconds since January 1, 1970 UTC
+ **/
```
pwithnall: Nitpick: I can’t be bothered to file suggestions for all of these, but they all need `Since: 0.2.1` and to use `*/` to terminate, rather than `**/` (the former is the gtk-doc preferred style).

libeos-payg/clock.c
```diff
@@ -0,0 +1,90 @@
...
+/**
+ * epg_clock_get_time:
+ * @self: an #EpgClock
+ *
+ * This returns the time based on the same clock used for
+ * epg_clock_source_new_seconds()
+ *
+ * Returns: the time since some unspecified starting point, in microseconds
+ **/
+gint64 epg_clock_get_time (EpgClock *self)
```
pwithnall:
```suggestion
gint64
epg_clock_get_time (EpgClock *self)
```

libeos-payg/clock.c
```diff
@@ -0,0 +1,90 @@
...
+/**
+ * epg_clock_source_new_seconds:
+ * @interval: the timeout interval in seconds
+ *
+ * Like g_timeout_source_new_seconds(), but the implementation determines what
+ * type of clock to use.
+ *
+ * Returns: the newly-created timeout source
+ **/
+GSource *epg_clock_source_new_seconds (EpgClock *self,
```
pwithnall:
```suggestion
GSource *
epg_clock_source_new_seconds (EpgClock *self,
```

libeos-payg/clock.c
```diff
@@ -0,0 +1,90 @@
...
+}
+
+/**
+ * epg_clock_source_new_seconds:
+ * @interval: the timeout interval in seconds
+ *
+ * Like g_timeout_source_new_seconds(), but the implementation determines what
+ * type of clock to use.
+ *
+ * Returns: the newly-created timeout source
```
pwithnall:
```suggestion
 * Returns: (transfer full): the newly-created timeout source
```

libeos-payg/real-clock.c
```diff
@@ -0,0 +1,85 @@
...
+#include <boottime-source.h>
+
+#define MSEC_PER_SEC 1000
+
+struct _EpgRealClock
+{
+  GObject parent_instance;
+};
+
+gint64 epg_real_clock_get_wallclock_time (EpgClock *clock)
```
pwithnall:
```suggestion
gint64
epg_real_clock_get_wallclock_time (EpgClock *clock)
```

libeos-payg/real-clock.c
```diff
@@ -0,0 +1,85 @@
...
+ * which case the provisions of the LGPL are applicable instead of those above.
+ * If you wish to allow use of your version of this file only under the terms
+ * of the LGPL, and not to allow others to use your version of this file under
+ * the terms of the MPL, indicate your decision by deleting the provisions
+ * above and replace them with the notice and other provisions required by the
+ * LGPL. If you do not delete the provisions above, a recipient may use your
+ * version of this file under the terms of either the MPL or the LGPL.
+ */
+
+#include <real-clock.h>
```
pwithnall:
```suggestion
#include <libeos-payg/real-clock.h>
```

libeos-payg/real-clock.c
```diff
@@ -0,0 +1,85 @@
...
+#include <boottime-source.h>
```
pwithnall:
```suggestion
#include <libeos-payg/boottime-source.h>
```

libeos-payg/real-clock.c
```diff
@@ -0,0 +1,85 @@
...
+gint64 epg_real_clock_get_wallclock_time (EpgClock *clock)
+{
+  EpgRealClock *self = (EpgRealClock *)clock;
+
+  g_return_val_if_fail (EPG_IS_REAL_CLOCK (self), -1);
+
+  return g_get_real_time ();
+}
+
+gint64 epg_real_clock_get_time (EpgClock *clock)
```
pwithnall:
```suggestion
gint64
epg_real_clock_get_time (EpgClock *clock)
```

libeos-payg/real-clock.c
```diff
@@ -0,0 +1,85 @@
...
+{
+  EpgRealClock *self = (EpgRealClock *)clock;
+
+  g_return_val_if_fail (EPG_IS_REAL_CLOCK (self), -1);
+
+  return epg_get_boottime ();
+}
+
+GSource *epg_real_clock_source_new_seconds (EpgClock *clock,
```
pwithnall:
```suggestion
GSource *
epg_real_clock_source_new_seconds (EpgClock *clock,
```

libeos-payg/real-clock.h
```diff
@@ -0,0 +1,39 @@
...
+#include <glib.h>
+
+G_BEGIN_DECLS
+
+#define EPG_TYPE_REAL_CLOCK epg_real_clock_get_type ()
+G_DECLARE_FINAL_TYPE (EpgRealClock, epg_real_clock, EPG, REAL_CLOCK, GObject)
+
+gint64 epg_real_clock_get_wallclock_time (EpgClock *self);
+
+gint64 epg_real_clock_get_time (EpgClock *self);
```
pwithnall: Any reason you’ve made these virtual function implementations public? Surely callers should just call `epg_clock_get_time()` instead?

libeos-payg/manager.c
```diff
@@ -401,6 +419,7 @@ epg_manager_set_property (GObject      *object,
  *    see #EpgManager:key-file
  * @state_directory: (transfer none) (optional): directory to load/store state
  *    in, or %NULL to use the default directory; see #EpgManager:state-directory
+ * @clock: an #EpgClock
```
pwithnall:
```suggestion
 * @clock: (transfer none): an #EpgClock
```

libeos-payg/manager.c
```diff
@@ -938,15 +974,35 @@ epg_manager_init_async (GAsyncInitable      *initable,
...
-                              file_load_cb, g_object_ref (task));
+      g_file_load_contents_async (wallclock_time_file, cancellable,
+                                  file_load_cb, g_object_ref (task));
+
+      /* Load the expiry seconds. */
+      g_autoptr(GFile) expiry_seconds_file = get_expiry_seconds_file (self);
+
+      g_file_load_contents_async (expiry_seconds_file, cancellable,
+                                  file_load_cb, g_object_ref (task));
+    }
```
pwithnall: I’d do this the other way round, and try loading the wallclock time file first. If that fails, fall back to the expiry time file. Don’t use `g_file_query_exists()` at all — just handle `G_IO_ERROR_NOT_FOUND` from the file loading functions. That saves on syscalls as well as eliminating the race condition.

libeos-payg/manager.c
```diff
@@ -1017,10 +1111,51 @@ file_load_cb (GObject      *source_object,
...
+          set_expiry_time (self, now_secs, 0);
+        }
+      else
+        {
+          union
+            {
+              guint64 u64;
+              gchar u8[8];
+            } expiry_secs;
+          G_STATIC_ASSERT (sizeof (expiry_secs.u8) == sizeof (self->expiry_time_secs));
```
pwithnall: Looks like the code for reading a 64-bit int from a file could be factored out.

libeos-payg/manager.c
```diff
@@ -1017,10 +1111,51 @@ file_load_cb (GObject      *source_object,
...
+           * want self->expiry_time_secs to be in terms of CLOCK_BOOTTIME, so
+           * we're migrating here */
           memcpy (expiry_time_secs.u8, data, sizeof (expiry_time_secs.u8));
           set_expiry_time (self, now_secs,
-                           (expiry_time_secs.u64 > now_secs) ? expiry_time_secs.u64 - now_secs : 0);
+                           (expiry_time_secs.u64 > wallclock_now_secs) ? expiry_time_secs.u64 - wallclock_now_secs : 0);
+
+          /* Delete the file since we now use clock-time and expiry-seconds instead */
+          g_file_delete_async (file, G_PRIORITY_DEFAULT, cancellable,
+                               file_load_delete_cb, g_object_ref (task));
```
pwithnall: Don’t you need to call `epg_multi_task_increment()` (IIRC) if you’re adding another async call to the chain?

libeos-payg/manager.c
```diff
@@ -88,6 +88,12 @@ static void        shutdown_save_state_cb (GObject      *source_object,
 static guint64     epg_manager_get_expiry_time     (EpgProvider *provider);
 static gboolean    epg_manager_get_enabled         (EpgProvider *provider);
 static guint64     epg_manager_get_rate_limit_end_time (EpgProvider *provider);
+static EpgClock *  epg_manager_get_clock (EpgProvider *provider);
+
+static guint64 last_save_time; /* wallclock timestamp of last state save */
+static guint64 last_save_expiry; /* seconds left to expiration at time of last state save */
+static gboolean last_save_time_set = FALSE; /* whether last_save_time has been set */
+static gboolean last_save_expiry_set = FALSE; /* whether last_save_expiry has been set */
```
pwithnall: Move these static variables inside the `EpgManager` object, or there’s no point in having an object.

pwithnall:
There’s a `g_get_real_time()` call in `epg_manager_service_manager_add_code()`. Does that also need to be changed to use the new `EpgClock`?

pwithnall:
Also, the `CLOCK_BOOTTIME` stuff is something we could potentially want in GLib. Fancy upstreaming it, or filing a task to do that in future?

libeos-payg/service.c
```diff
@@ -229,8 +231,9 @@ provider_get_first_enabled_cb (GObject      *source_object,
       return;
     }
 
+  clock = epg_real_clock_new ();
```
wjt: Won't you also need to feed this into `epg_provider_loader_get_first_enabled_async`?

libeos-payg/manager.c
```diff
@@ -647,7 +652,7 @@ set_expiry_time (EpgManager *self,
     {
       g_assert (self->expiry_time_secs >= now_secs);
       g_assert (self->expiry_time_secs - now_secs <= G_MAXUINT);
-      self->expiry = g_timeout_source_new_seconds (MIN (self->expiry_time_secs - now_secs, 60));
+      self->expiry = epg_clock_source_new_seconds (self->clock, MIN (self->expiry_time_secs - now_secs, 60));
```
wjt: I think the point of ticking the clock every 60 seconds was to ensure that changes to the wall-clock time were noticed within 60 seconds? So in principle this could actually just be set to seconds-to-expiry.

libeos-payg/manager.c
```diff
@@ -954,15 +974,35 @@ epg_manager_init_async (GAsyncInitable      *initable,
 
   EpgManager *self = EPG_MANAGER (initable);
   g_autoptr(GTask) task = g_task_new (self, cancellable, callback, user_data);
+  g_autoptr(GFile) expiry_time_file = get_expiry_time_file (self);
+
   g_task_set_source_tag (task, epg_manager_init_async);
   g_task_set_priority (task, priority);
   epg_multi_task_attach (task, 4);
```
wjt: In the case where you're making 1 extra async call, this will need incrementing.

libeos-payg/manager.c
```diff
@@ -1110,6 +1229,32 @@ file_load_cb (GObject      *source_object,
...
+        {
+          /* Time has gone backwards!? Either the saved time is wrong (and
+           * there's no way to know by how much) or the current time is wrong
+           * (and NTP will fix it, see T24501). Let's just assume time stood
+           * still. */
+          set_expiry_time (self, now_secs, last_save_expiry);
+        }
+      else
+        {
+          /* Time has continued its inexorable march forward while the computer
```
wjt: Very good

libeos-payg/manager.c
```diff
@@ -1121,19 +1266,27 @@ file_load_delete_cb (GObject      *source_object,
...
+      local_error = g_error_new (G_IO_ERROR, G_IO_ERROR_INVALID_DATA,
+                                 _("State file ‘%s’ was the wrong length."),
+                                 file_path);
+    }
 
-  g_clear_error (&local_error);
-  local_error = g_error_new (G_IO_ERROR, G_IO_ERROR_INVALID_DATA,
-                             _("State file ‘%s’ was the wrong length."),
-                             file_path);
   epg_multi_task_return_error (task, G_STRFUNC, g_steal_pointer (&local_error));
```
wjt: Does this do the right thing (ie, possibly return success) if passed `NULL`?

libeos-payg/manager.c
wjt: In -nonfree I used `GData[Input|Output]Stream` which might be worth a look.

libeos-payg/real-clock.c
```diff
@@ -0,0 +1,85 @@
...
+                                            guint     interval)
+{
+  EpgRealClock *self = (EpgRealClock *)clock;
+  guint interval_clamped;
+
+  g_return_val_if_fail (EPG_IS_REAL_CLOCK (self), NULL);
+
+  interval_clamped = MIN (interval, G_MAXUINT / MSEC_PER_SEC);
```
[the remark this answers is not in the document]
pwithnall: Sigh. Looks like that needs an upstream fix. :-(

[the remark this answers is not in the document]
wjt: https://gitlab.gnome.org/GNOME/glib/merge_requests/496

libeos-payg/boottime-source.c
mwleeds: With that we have to cast it when using it in `g_source_new` so it doesn't seem worth it (even though `g_source_new` seems to treat it as `const`).

libeos-payg/manager.c
mwleeds: Yes I think you're right, and similarly for the existing `g_file_delete_async` calls

libeos-payg/manager.c
mwleeds: You're right, that seems better. It's slightly unfortunate that it means we can't have a clean separation between loading files in the `init_async` and processing them in the callback, but that's a small price to pay.

mwleeds:
> There’s a g_get_real_time() call in epg_manager_service_manager_add_code(). Does that also need to be changed to use the new EpgClock?

That will be addressed by the one of the remaining tasks for this branch, removing the `now` parameter from `epg_provider_add_code`

mwleeds:
> Also, the CLOCK_BOOTTIME stuff is something we could potentially want in GLib. Fancy upstreaming it, or filing a task to do that in future?

I put this on a todo list but I can't promise when it will happen. Or if you or Will get around to it first that works too. Here's a task: https://phabricator.endlessm.com/T24648

libeos-payg/boottime-source.c
```diff
@@ -0,0 +1,154 @@
...
+
+  its.it_value = its.it_interval;
+
+  fd = timerfd_create (CLOCK_BOOTTIME, TFD_CLOEXEC);
+  if G_UNLIKELY (fd < 0)
+    g_error ("timerfd_create (CLOCK_BOOTTIME, TFD_CLOEXEC) failed: %s",
+             g_strerror (errno));
+
+  if G_UNLIKELY (timerfd_settime (fd,
+                                  0 /*TFD_NONBLOCK */,
```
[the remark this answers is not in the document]
mwleeds: So we should pass it to `timerfd_create` then?

libeos-payg/service.c
mwleeds: Good catch. I opted to instead change `epg_manager_constructed` to initialize the `clock` member with an `EpgRealClock`, which matches how the other construct-time properties are handled.

libeos-payg/manager.c
mwleeds: Yes I think you're right

libeos-payg/real-clock.h
mwleeds: Yes you're right

libeos-payg/manager.c
mwleeds: I ended up going the safe route and using helper functions but otherwise leaving the code untouched.

mwleeds:
I've now responded to all the feedback with either a fixup commit or a response (or both). This branch is still WIP though with this todo list:

- [ ] Remove the now parameter in epg_provider_add_code
- [ ] Update the unit tests to use a fake clock

libeos-payg/boottime-source.c
pwithnall: It’s worth it because it documents your intent for the structure, and (without checking the assembly) means the compiler can put it in a read-only section of the object file.

libeos-payg/boottime-source.c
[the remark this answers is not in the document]
pwithnall: Yes.

libeos-payg/clock.c
```diff
@@ -0,0 +1,95 @@
...
+ * version of this file under the terms of either the MPL or the LGPL.
+ */
+
+#include <clock.h>
+#include <glib.h>
+
+G_DEFINE_INTERFACE (EpgClock, epg_clock, G_TYPE_OBJECT)
+
+static void
+epg_clock_default_init (EpgClockInterface *iface)
```
pwithnall: Add `gpointer iface_data` as in my other comment.

libeos-payg/fake-clock.c
```diff
@@ -0,0 +1,147 @@
...
+static void
+clock_iface_init (EpgClockInterface *iface)
+{
+  iface->get_wallclock_time = epg_fake_clock_get_wallclock_time;
+  iface->get_time = epg_fake_clock_get_time;
+  iface->source_new_seconds = epg_fake_clock_source_new_seconds;
+}
+
+G_DEFINE_TYPE_WITH_CODE (EpgFakeClock, epg_fake_clock, G_TYPE_OBJECT,
+                         G_IMPLEMENT_INTERFACE (EPG_TYPE_FAKE_CLOCK, clock_iface_init))
```
pwithnall: This should be `EPG_TYPE_CLOCK`, not `EPG_TYPE_FAKE_CLOCK`. That’s what was causing the `epg_fake_clock_get_type()` recursion and hence a deadlock:
```
Thread 1 (Thread 0x7ffff6ba0380 (LWP 2587)):
#0  0x00007ffff751aead in syscall () from /lib64/libc.so.6
#1  0x00007ffff76c7160 in g_cond_wait (cond=0x7ffff7949560 <g_once_cond>, mutex=0x7ffff7949550 <g_once_mutex>) at /opt/endless/source/glib/glib/gthread-posix.c:1395
#2  0x00007ffff769fab8 in g_once_init_enter (location=0x7ffff7fc39d0 <g_define_type_id.volatile>) at /opt/endless/source/glib/glib/gthread.c:665
#3  0x00007ffff7f9e51e in epg_fake_clock_get_type () at ../../source/eos-payg/libeos-payg/fake-clock.c:116
#4  0x00007ffff7f9e4aa in epg_fake_clock_get_type_once () at ../../source/eos-payg/libeos-payg/fake-clock.c:116
#5  0x00007ffff7f9e552 in epg_fake_clock_get_type () at ../../source/eos-payg/libeos-payg/fake-clock.c:116
#6  0x00007ffff7f9e5f4 in epg_fake_clock_new () at ../../source/eos-payg/libeos-payg/fake-clock.c:146
#7  0x000000000040322e in manager_new_failable (fixture=0x41ee20, enabled=1, error=0x7fffffffc820) at ../../source/eos-payg/libeos-payg/tests/manager.c:166
#8  0x0000000000403368 in manager_new (fixture=0x41ee20) at ../../source/eos-payg/libeos-payg/tests/manager.c:181
#9  0x000000000040396b in test_manager_load_empty (fixture=0x41ee20, data=0x0) at ../../source/eos-payg/libeos-payg/tests/manager.c:241
#10 0x00007ffff769cb1f in test_case_run (tc=0x41fa90) at /opt/endless/source/glib/glib/gtestutils.c:2163
#11 0x00007ffff769cedf in g_test_run_suite_internal (suite=0x41f040, path=0x0) at /opt/endless/source/glib/glib/gtestutils.c:2247
#12 0x00007ffff769cf88 in g_test_run_suite_internal (suite=0x41f020, path=0x0) at /opt/endless/source/glib/glib/gtestutils.c:2259
#13 0x00007ffff769d19f in g_test_run_suite (suite=0x41f020) at /opt/endless/source/glib/glib/gtestutils.c:2335
#14 0x00007ffff769be2e in g_test_run () at /opt/endless/source/glib/glib/gtestutils.c:1600
#15 0x0000000000406b73 in main (argc=1, argv=0x7fffffffcc68) at ../../source/eos-payg/libeos-payg/tests/manager.c:789
```

mwleeds: That makes sense; I think it was a replace-all error. Thanks for pointing it out!

libeos-payg/fake-clock.c
```diff
@@ -0,0 +1,147 @@
...
+  guint interval_clamped;
+
+  g_return_val_if_fail (EPG_IS_FAKE_CLOCK (self), NULL);
+
+  interval_clamped = MIN (interval, G_MAXUINT / MSEC_PER_SEC);
+  return epg_boottime_source_new (interval_clamped * MSEC_PER_SEC);
+}
+
+static void
+clock_iface_init (EpgClockInterface *iface)
```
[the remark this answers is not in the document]
mwleeds: Okay. I wonder why none of the interfaces in gnome-builder seem to use `iface_data`, and why I'm not seeing that warning

libeos-payg/clock.c
mwleeds: If I do that I get
```
../libeos-payg/clock.c:26:1: error: conflicting types for ‘epg_clock_default_init’
 epg_clock_default_init (EpgClockInterface *iface,
 ^~~~~~~~~~~~~~~~~~~~~~
In file included from /usr/include/glib-2.0/gobject/gobject.h:24:0,
                 from /usr/include/glib-2.0/gobject/gbinding.h:29,
                 from /usr/include/glib-2.0/glib-object.h:23,
                 from ../libeos-payg/clock.h:22,
                 from ../libeos-payg/clock.c:20:
../libeos-payg/clock.c:23:31: note: previous declaration of ‘epg_clock_default_init’ was here
 G_DEFINE_INTERFACE (EpgClock, epg_clock, G_TYPE_OBJECT)
                               ^
/usr/include/glib-2.0/gobject/gtype.h:2008:17: note: in definition of macro ‘_G_DEFINE_INTERFACE_EXTENDED_BEGIN’
 static void     type_name##_default_init        (TypeName##Interface *klass); \
                 ^~~~~~~~~
/usr/include/glib-2.0/gobject/gtype.h:1756:47: note: in expansion of macro ‘G_DEFINE_INTERFACE_WITH_CODE’
 #define G_DEFINE_INTERFACE(TN, t_n, T_P)      G_DEFINE_INTERFACE_WITH_CODE(TN, t_n, T_P, ;)
                                               ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
../libeos-payg/clock.c:23:1: note: in expansion of macro ‘G_DEFINE_INTERFACE’
 G_DEFINE_INTERFACE (EpgClock, epg_clock, G_TYPE_OBJECT)
 ^~~~~~~~~~~~~~~~~~
```

mwleeds: Maybe the difference is that it's an interface that inherits from `GOBject` not `GAsyncInitable`

mwleeds: Surely there's documentation about this somewhere but I haven't been able to find it

libeos-payg/fake-clock.c
treitter: Called it :)

libeos-payg/fake-clock.c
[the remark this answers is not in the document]
pwithnall: You need to be using GCC 8, `-Wcast-function-type`, and an old version of GLib which doesn’t have https://gitlab.gnome.org/GNOME/glib/commit/c1f5e52844c3bf503aca8881b6af68f78f028b4d.

Relevant: https://gitlab.gnome.org/GNOME/glib/merge_requests/512

libeos-payg/clock.c
pwithnall: Interesting. Maybe just leave out the `iface_data` arguments for now. I’ll investigate this new failure from the GLib side at some point.

> Surely there's documentation about this somewhere but I haven't been able to find it

lol no

libeos-payg/fake-clock.c
pwithnall: 🖁

libeos-payg/fake-clock.c
```diff
@@ -0,0 +1,167 @@
...
+
+static gint64
+epg_fake_clock_get_time (EpgClock *clock)
+{
+  EpgFakeClock *self = (EpgFakeClock *)clock;
+
+  g_return_val_if_fail (EPG_IS_FAKE_CLOCK (self), -1);
+
+  if (self->time == 0)
+    self->time = epg_get_boottime () / G_USEC_PER_SEC;
```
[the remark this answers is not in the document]
mwleeds: The idea was that the tests should use realistic values to test realistically. You could still use an explicit value in a test if you want, by either passing it to `epg_fake_clock_new` or to `set_time`/`set_wallclock_time`. Does it matter if the unit tests use slightly different values each time?

libeos-payg/boottime-source.c
```diff
@@ -0,0 +1,159 @@
...
+epg_get_boottime (void)
+{
+  struct timespec ts;
+  gint result;
+
+  result = clock_gettime (CLOCK_BOOTTIME, &ts);
+
+  if (G_UNLIKELY (result != 0))
+    g_error ("clock_gettime (CLOCK_BOOTTIME) failed: %s",
+             g_strerror (errno));
```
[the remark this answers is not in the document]
mwleeds: Yeah I don't think so, unless it's easy and you already know how

libeos-payg/clock.c
```diff
@@ -0,0 +1,95 @@
...
+  g_assert (iface->get_time != NULL);
+  return iface->get_time (self);
+}
+
+/**
+ * epg_clock_source_new_seconds:
+ * @interval: the timeout interval in seconds
+ *
+ * Like g_timeout_source_new_seconds(), but the implementation determines what
+ * type of clock to use.
```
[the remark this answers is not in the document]
mwleeds: The GSource documentation says "For timeout sources, the prepare and check functions both return TRUE if the timeout interval has expired." So I think that means we can say the GSource will fire the next time it's checked for a zero interval. I wouldn't expect a future implementation to need to differ from that.

libeos-payg/manager.c
```diff
@@ -962,6 +1001,23 @@ epg_manager_init_async (GAsyncInitable      *initable,
   epg_multi_task_return_boolean (task, TRUE);
 }
 
+/* @bytes must have length 8 */
+static guint64
+get_guint64_from_bytes (gchar *bytes)
```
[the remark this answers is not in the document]
mwleeds: There's a note about endianness in the `EpgManager` docs, but sure I'll copy it

libeos-payg/manager.c
```diff
@@ -1152,6 +1320,32 @@ static void file_replace_cb     (GObject      *source_object,
...
+                       GCancellable *cancellable)
+{
+  union
+    {
+      guint64 u64;
+      const guint8 u8[8];
+    } number_union;
+  number_union.u64 = number;
+
+  g_autoptr(GBytes) number_bytes = g_bytes_new (number_union.u8,
```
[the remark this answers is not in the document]
mwleeds: Strangely when I implemented this change it broke one of the unit tests so I reverted it.

libeos-payg/boottime-source.c
```diff
@@ -0,0 +1,159 @@
...
+      .tv_sec = interval_ms / 1000,
+      .tv_nsec = (interval_ms % 1000) * (1000 * 1000),
+    }
+  };
+
+  its.it_value = its.it_interval;
+
+  fd = timerfd_create (CLOCK_BOOTTIME, TFD_CLOEXEC | TFD_NONBLOCK);
+  if (G_UNLIKELY (fd < 0))
+    g_error ("timerfd_create (CLOCK_BOOTTIME, TFD_CLOEXEC) failed: %s",
```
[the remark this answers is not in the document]
pwithnall: > if the system hit the limit of open file descriptors it's probably unusable anyway, right?

Not really. I’ve hit the limit with Builder before (I think it leaked at once point) and the system is still fairly usable. You can’t start any new applications, but old ones generally continue to work fine, and once you’ve closed the application which is leaking them all, the system recovers well without a restart. We don’t want to make that situation any worse.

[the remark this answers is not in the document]
pwithnall: Put another way: running out of FDs is not a programmer error, so generally shouldn’t be handled with `g_error()`.

libeos-payg/manager.c
```diff
@@ -647,7 +669,7 @@ set_expiry_time (EpgManager *self,
  * would overflow, set the expiry time as high as possible. Everything is
  * handled in seconds.
  *
- * @now will typically be the value of (g_get_real_time() / G_USEC_PER_SEC).
+ * @now will typically be the current value of CLOCK_BOOTTIME.
  * It is parameterised to allow easy testing. */
```
wjt: This comment isn't really true because this function is static. The function that used to have the same parameter for ease of testing no longer does.

libeos-payg/manager.c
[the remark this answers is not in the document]
wjt: Probably because the buffer passed to `g_file_replace_contents_async` must remain valid until its callback is called, but this is a stack variable that will become invalid almost immediately.

libeos-payg/boottime-source.c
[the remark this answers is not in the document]
pwithnall: It would be something like the following in the top-level `meson.build` (untested):
```
code = '''#include <time.h>
void func (void) { struct timespec ts; clock_gettime (CLOCK_BOOTTIME, &ts); }
'''
if not cc.compiles(code, name : 'CLOCK_BOOTTIME is supported')
  error('clock_gettime(CLOCK_BOOTTIME) is not supported')
endif
```

libeos-payg/manager.c
```diff
@@ -1166,30 +1360,20 @@ epg_manager_save_state_async (EpgProvider         *provider,
...
-                                       NULL,  /* ETag */
-                                       FALSE,  /* no backup */
-                                       G_FILE_CREATE_PRIVATE,
-                                       cancellable,
-                                       file_replace_cb,
-                                       g_object_ref (task));
+  /* Save the expiry seconds. */
+  g_autoptr(GFile) expiry_seconds_file = get_expiry_seconds_file (self);
+  guint64 now_secs = epg_clock_get_time (self->clock);
+  if (now_secs > self->expiry_time_secs)
```
wjt: There's a lot of `x > y ? 0 : y - x` in this diff (whether literally, or as here split across several statements). I suppose `MAX (0, y - x)` would be considered Very Bad in the case of underflow, though.

libeos-payg/fake-clock.c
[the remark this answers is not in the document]
pwithnall: You can use realistic values and still have the tests be deterministic, for example, by choosing static clock values from 2018 rather than the bottom of the range of valid integers.

> Does it matter if the unit tests use slightly different values each time?

As a general principle, yes, this makes for hard to reproduce failures in future. If the test inputs change and the code under test changes, then the test fails — what’s caused the failure?

libeos-payg/service.c
wjt: If you're relying on that defaulting behaviour, why not drop this and pass `NULL`?

libeos-payg/manager.c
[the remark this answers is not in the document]
pwithnall: Or refer to it.

libeos-payg/manager.c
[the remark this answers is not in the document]
pwithnall: Do you know how it broke? Would be worth investigating a bit.

libeos-payg/tests/manager.c
```diff
@@ -317,6 +325,21 @@ test_manager_code_format_rejects (Fixture *fixture,
   g_assert_false (g_match_info_is_partial_match (match_info));
 }
 
+static void
+write_valid_clock_time_file (Fixture *fixture)
+{
+  guint64 time = g_get_real_time () / G_USEC_PER_SEC;
```
wjt: I'm not a fan of variables that shadow standard-library function names

libeos-payg/fake-clock.c
```diff
@@ -0,0 +1,167 @@
...
+epg_fake_clock_init (EpgFakeClock *self)
+{
+}
+
+/**
+ * epg_fake_clock_new:
+ * @optional_time: A positive time value, or -1
+ * @optional_wallclock_time: A positive wallclock time value, or -1
+ *
+ * EpgFakeClock is a clock that can be arbitrarily set forward or backward, and
```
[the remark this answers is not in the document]
pwithnall: I think it’s possible for the wall clock time to go backwards if someone changes the system clock, right? We should probably have the flexibility to test for that, even if we are relying on lower layers in the system preventing the user from changing the system clock.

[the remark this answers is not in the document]
wjt: True. Wallclock yes, the other one no.

libeos-payg/clock.c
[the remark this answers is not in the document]
mwleeds: Added a fixup commit

libeos-payg/fake-clock.c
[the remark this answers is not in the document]
mwleeds: I implemented this in a fixup

libeos-payg/manager.c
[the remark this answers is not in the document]
mwleeds: Yeah I think Will is right about the reason it didn't work to use `g_file_replace_contents_async`. Anyway, I pushed a fixup with a simpler implementation.

libeos-payg/service.c
mwleeds: Pushed a fixup

libeos-payg/boottime-source.c
```diff
@@ -88,18 +89,21 @@ static const GSourceFuncs epg_boottime_source_funcs = {
...
  *
  * Like g_timeout_source_new(), but uses `CLOCK_BOOTTIME` to account for time
  * when the system is suspended.
  *
  * If @interval_ms is set to zero, the #GSource will be ready the next time
  * it's checked.
  *
- * Returns: (transfer full): a new `CLOCK_BOOTTIME` #GSource
+ * Returns: (transfer full) (nullable): a new `CLOCK_BOOTTIME` #GSource, or
+ *   %NULL with @error set
```
pwithnall: Don’t add the `(nullable)` annotation for the case where a function can return `NULL` iff it returns an error. The existence of a `GError**` argument implies that. Only add `(nullable)` if it’s possible for a function to return `NULL` _without_ an error set.

Would also be useful to document the error code and the situation it can be returned in, i.e.
```
@error will be set to a #GIOError if, for example, the process runs out of file descriptors.
```

libeos-payg/boottime-source.c
```diff
@@ -113,11 +117,19 @@ epg_boottime_source_new (guint interval_ms)
...
   its.it_value = its.it_interval;
 
+  /* Set the GError if timerfd_create() fails because it could be e.g. ENFILE
+   * which we should handle gracefully */
   fd = timerfd_create (CLOCK_BOOTTIME, TFD_CLOEXEC | TFD_NONBLOCK);
   if (G_UNLIKELY (fd < 0))
-    g_error ("timerfd_create (CLOCK_BOOTTIME, TFD_CLOEXEC) failed: %s",
-             g_strerror (errno));
+    {
+      g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED,
```
pwithnall: Use `g_io_error_from_errno()`. I wish `G_IO_ERROR_FAILED` had never been invented; people use it too much when more specific errors are available.

libeos-payg/tests/boottime-source.c
```diff
@@ -40,11 +40,16 @@ timeout_cb (gpointer data)
...
 {
-  g_autoptr(GSource) source = epg_boottime_source_new (INTERVAL_MS);
+  g_autoptr(GError) local_error = NULL;
+  g_autoptr(GSource) source = NULL;
   guint called = 0;
   gint64 start = epg_get_boottime ();
   gint64 end;
 
+  source = epg_boottime_source_new (INTERVAL_MS, &local_error);
+  g_assert (source != NULL);
```
pwithnall: `g_assert_nonnull()`, and do the `g_assert_no_error()` call first to get more detailed error messages out on failure.

libeos-payg/tests/boottime-source.c
```diff
@@ -62,11 +67,16 @@ test_once (void)
...
```
pwithnall: `g_assert_nonnull()`

libeos-payg/clock.c
```diff
@@ -75,22 +75,26 @@ epg_clock_get_time (EpgClock *self)
  * epg_clock_source_new_seconds:
  * @self: an #EpgClock
  * @interval: the timeout interval in seconds
+ * @error: return location for a #GError, or %NULL
  *
  * Like g_timeout_source_new_seconds(), but the implementation determines what
  * type of clock to use.
  *
- * Returns: (transfer full): the newly-created timeout source
+ * Returns: (transfer full) (nullable): the newly-created timeout source, or
```
pwithnall: Drop `(nullable)`

libeos-payg/manager.c
```diff
@@ -1336,29 +1336,17 @@ static void file_save_delete_cb (GObject      *source_object,
...
-                                       NULL,  /* ETag */
-                                       FALSE,  /* no backup */
-                                       G_FILE_CREATE_PRIVATE,
-                                       cancellable,
-                                       file_replace_cb,
-                                       g_object_ref (task));
+  g_autofree gchar *file_path = g_file_get_path (file);
+  g_autoptr(GError) local_error = NULL;
+
+  if (!g_file_set_contents (file_path, (char *)&number, sizeof(number), &local_error))
```
pwithnall: Not to be That Guy again, but this is synchronous and will block the main thread.

libeos-payg/fake-clock.c
```diff
@@ -106,18 +106,82 @@ epg_fake_clock_set_time (EpgFakeClock *self,
...
 
   g_return_val_if_fail (EPG_IS_FAKE_CLOCK (self), NULL);
 
-  interval_clamped = MIN (interval, G_MAXUINT / MSEC_PER_SEC);
-  return epg_boottime_source_new (interval_clamped * MSEC_PER_SEC, error);
+  source = g_source_new ((GSourceFuncs *)&epg_fake_source_funcs,
+                         sizeof (EpgFakeSource));
+  fake_source = (EpgFakeSource *)source;
+  fake_source->clock = g_object_ref (self);
+  fake_source->ready_time = epg_clock_get_time (EPG_CLOCK (self)) + interval;
```
pwithnall: Nitpick: Would be good to put in an assertion to make sure this addition doesn’t overflow.

libeos-payg/fake-clock.c
```diff
@@ -106,18 +106,82 @@ epg_fake_clock_set_time (EpgFakeClock *self,
   self->time_secs = new_time;
 }
 
+typedef struct {
+  GSource parent;
+
+  EpgFakeClock *clock; /* owned */
+  gint64 ready_time;
```
pwithnall: Needs renaming or a comment to indicate what its units are.

libeos-payg/manager.c
```diff
@@ -299,6 +311,9 @@ epg_manager_constructed (GObject *object)
 
   if (self->state_directory == NULL)
     self->state_directory = g_file_new_for_path (LOCALSTATEDIR "/lib/eos-payg");
+
+  if (self->clock == NULL)
+    self->clock = (EpgClock *)epg_real_clock_new ();
```
wjt:
```suggestion
    self->clock = EPG_CLOCK (epg_real_clock_new ());
```

libeos-payg/manager.c
```diff
@@ -617,23 +644,32 @@ set_expiry_time (EpgManager *self,
...
-  /* Set the expiry timer. g_timeout_source_new_seconds() takes a #guint, and
+  /* Set the expiry timer. epg_clock_source_new_seconds() takes a #guint, and
    * @span is a #guint64. However, the maximum span is 365 days, which is
-   * representable in 32 bits. We don’t set a timer for infinite periods.
-   *
-   * FIXME: For the moment, poll every 60s until the expiry time is reached, so
-   * we don’t have to worry about recalculating the timeout period after
-   * resuming from suspend, or if the system clock or timezone changes.
-   * See: https://phabricator.endlessm.com/T22074 */
+   * representable in 32 bits. We don’t set a timer for infinite periods. */
```
wjt: I know this is existing code but it's not actually true that the maximum time to expiry is 365 days. The maximum period for a code is indeed 365 days, or 2 ** 24.910496345705575 seconds, but there are 2 ** 8 possible codes for each period, and @treitter is working away at making it easy to add a code before the old one has expired. If you add 137 365-day codes, the expiry time is more than 2 ** 32 seconds into the future, so the assertion below will fail.

libeos-payg/manager.c
```diff
@@ -1060,6 +1209,8 @@ file_load_cb (GObject      *source_object,
 
               if (!epc_period_validate (used_code->period, &local_error))
                 {
+                  epg_multi_task_increment (task);
+                  g_task_set_task_data (task, g_steal_pointer (&invalid_data_error), (GDestroyNotify)g_error_free);
```
wjt: This is pretty ugly; I suppose it's my fault for making `epg_multi_task_return_error()` return immediately rather than waiting for all the other operations to complete. But would it be so bad to call it here?

libeos-payg/manager.c
mwleeds: Added a fixup commit to clamp to G_MAXUINT

libeos-payg/manager.c
mwleeds: Added a fixup

libeos-payg/manager.c
mwleeds: I think it's better to keep the pattern consistent of returning from a task in the callback after the file is written/deleted. Otherwise I don't understand how you'd know in `file_load_delete_cb` whether you need to return or if an error has already been returned to the task.

pwithnall:
I have no more comments. I think @wjt needs to reply to/resolve his remaining comments.

libeos-payg/clock.c
pwithnall: > > Surely there's documentation about this somewhere but I haven't been able to find it
> 
> lol no

(Finally) followed up upstream as https://gitlab.gnome.org/GNOME/glib/merge_requests/564.

mwleeds:
Merging this despite the build failure because it's unrelated and this should still build on eos3.5

PR #29, opened 2018-11-28
eos-payg-ctl: add --bus-address argument; tweak code style

wjt:
`-a` and `--bus-address` matches the short and long names used by the daemon (courtesy of libgsystemservice), and is useful when running a development copy of the daemon on the session bus.

https://github.com/ambv/black is a reasonable (and reasonably popular) auto-formatter for Python code. In a few places, it disagrees with flake8's defaults; tweak flake8 to ignore these.

Perhaps in a future life I'll make flake8- and black-cleanliness a compile-time check.

Preliminary work for https://phabricator.endlessm.com/T23781

### DIFF
--- a/eos-payg-ctl/eos-payg-ctl
+++ b/eos-payg-ctl/eos-payg-ctl
@@ -24,12 +24,12 @@ import sys
 from dateutil import tz
 from gi.repository import GLib, Gio
 
-BUS_NAME = 'com.endlessm.Payg1'
-OBJECT_PATH = '/com/endlessm/Payg1'
-INTERFACE = 'com.endlessm.Payg1'
-ERROR_DOMAIN = 'com.endlessm.Payg1.Error'
+BUS_NAME = "com.endlessm.Payg1"
+OBJECT_PATH = "/com/endlessm/Payg1"
+INTERFACE = "com.endlessm.Payg1"
+ERROR_DOMAIN = "com.endlessm.Payg1.Error"
 
-DBUS_PROPERTIES_INTERFACE = 'org.freedesktop.DBus.Properties'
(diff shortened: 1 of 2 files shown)


eos-payg-ctl/eos-payg-ctl
```diff
@@ -85,52 +101,59 @@ def __exit_on_payg_error():
         # Gio.dbus_error_get_remote_error(e) does not modify e.message,
         # presumably because the Python-side GLib.GError is marshalled back to
         # a new temporary C-side GError.
-        message = e.message[e.message.index(remote_error):]
+        message = e.message[e.message.index(remote_error) :]
```
pwithnall: Really? I’ve never seen that formatting before.

wjt: Nor had I. https://github.com/ambv/black#slices has some discussion. The behaviour depends on how complicated the expression on either side of the colon is:

```
✔ ~ ➭ black --diff test.py 
--- test.py	2018-11-28 15:41:50.180384 +0000
+++ test.py	2018-11-28 15:42:07.442551 +0000
@@ -1,3 +1,3 @@
 x[1:]
-y[a.b:]
+y[a.b :]
```

I don't love it, but I don't hate it.

pwithnall: ¯\_(ツ)_/¯  I have no enthusiasm for arguing about this. I’m OK with whatever.

wjt: Me either – the reason I like Black is that it has no options :)

pwithnall:
Ship it?

PR #31, opened 2019-01-02
eos-payg: Return the amount of time added

mazen-asef:
Return the amount of time added into the add_code function
of the daemon.

https://phabricator.endlessm.com/T24128

### DIFF
--- a/libeos-payg/manager-service.c
+++ b/libeos-payg/manager-service.c
@@ -835,16 +835,17 @@ epg_manager_service_manager_add_code (EpgManagerService     *self,
                                       GDBusMethodInvocation *invocation)
 {
   g_autoptr(GError) local_error = NULL;
+  gint64 time_added = 0;
 
   const gchar *code_str;
   g_variant_get (parameters, "(&s)", &code_str);
 
-  epg_provider_add_code (self->provider, code_str, &local_error);
+  epg_provider_add_code (self->provider, code_str, &time_added, &local_error);
 
   if (local_error != NULL)
     g_dbus_method_invocation_return_gerror (invocation, local_error);
   else
-    g_dbus_method_invocation_return_value (invocation, NULL);
+    g_dbus_method_invocation_return_value (invocation, g_variant_new ("(x)", time_added));
 }
 
 static void
--- a/libeos-payg/manager.c
+++ b/libeos-payg/manager.c
@@ -58,6 +58,7 @@ static gboolean epg_manager_init_finish (GAsyncInitable  *initable,
 
 static gboolean    epg_manager_add_code   (EpgProvider  *provider,
                                            const gchar  *code_str,
+                                           gint64       *time_added,
                                            GError      **error);
 static gboolean    epg_manager_clear_code (EpgProvider  *provider,
                                            GError      **error);
@@ -714,7 +715,7 @@ set_expiry_time (EpgManager *self,
  * handled in seconds.
  *
  * @now_secs will typically be the current value of CLOCK_BOOTTIME. */
-static void
+static gint64
 extend_expiry_time (EpgManager *self,
                     guint64     now_secs,
                     EpcPeriod   period)
@@ -806,6 +807,8 @@ extend_expiry_time (EpgManager *self,
     }
 
   set_expiry_time (self, NULL, FALSE, now_secs, span_secs);
+
+  return MIN(span_secs, G_MAXINT64);
 }
 
 static gint
@@ -827,6 +830,7 @@ used_codes_sort_cb (gconstpointer a,
 static gboolean
 epg_manager_add_code (EpgProvider   *provider,
                       const gchar  *code_str,
+                      gint64       *time_added,
                       GError      **error)
 {
   EpgManager *self = EPG_MANAGER (provider);
@@ -835,6 +839,7 @@ epg_manager_add_code (EpgProvider   *provider,
 
   g_return_val_if_fail (EPG_IS_MANAGER (provider), FALSE);
   g_return_val_if_fail (code_str != NULL, FALSE);
+  g_return_val_if_fail (time_added != NULL, FALSE);
   g_return_val_if_fail (error == NULL || *error == NULL, FALSE);
 
   if (!check_enabled (self, error))
@@ -880,7 +885,7 @@ epg_manager_add_code (EpgProvider   *provider,
   g_array_sort (self->used_codes, used_codes_sort_cb);
 
   /* Extend the expiry time. */
-  extend_expiry_time (self, now_secs, period);
+  *time_added = extend_expiry_time (self, now_secs, period);
 
   /* Reset the rate limiting history, since the code was successful. */
   clear_rate_limiting (self);
--- a/libeos-payg/provider.c
+++ b/libeos-payg/provider.c
@@ -155,18 +155,23 @@ epg_provider_default_init (EpgProviderInterface *iface)
  * epg_provider_add_code:
  * @self: an #EpgProvider
  * @code_str: code to verify and add
+ * @time_added: return location for the span of seconds added by code_str.
  * @error: return location for a #GError
  *
  * Verify and add the given @code_str. This checks that @code_str is valid, and
  * has not been used already. If so, it will add the time period given in the
  * @code_str to #EpgProvider:expiry-time (or to the current time if
- * #EpgProvider:expiry-time is in the past). If @code_str fails verification or
- * cannot be added, an error will be returned.
+ * #EpgProvider:expiry-time is in the past). On success, @time_added will be
+ * populated by the span of seconds added by @code_str. If @code_str fails
+ * verification or cannot be added, an error will be returned.
  *
  * Calls to this function may be rate limited: if too many attempts are made within
  * a given time period, %EPG_MANAGER_ERROR_TOO_MANY_ATTEMPTS will be returned
  * until that period expires. The rate limiting history is reset on a successful
- * verification of a code.
+ * verification of a code. 
+ *
+ * @time_added is signed because some implementations can reduce the available credit
+ * when you add a new code.
  *
  * Returns: %TRUE on success, %FALSE otherwise
  * Since: 0.2.0
@@ -174,15 +179,17 @@ epg_provider_default_init (EpgProviderInterface *iface)
 gboolean
 epg_provider_add_code   (EpgProvider  *self,
                          const gchar  *code_str,
+                         gint64       *time_added,
                          GError      **error)
 {
   g_return_val_if_fail (EPG_IS_PROVIDER (self), FALSE);
+  g_return_val_if_fail (time_added != NULL, FALSE);
 
   EpgProviderInterface *iface = EPG_PROVIDER_GET_IFACE (self);
 
   g_assert (iface->add_code != NULL);
 
-  return iface->add_code (self, code_str, error);
+  return iface->add_code (self, code_str, time_added, error);
 }
 
 /**
--- a/libeos-payg/provider.h
+++ b/libeos-payg/provider.h
@@ -36,6 +36,7 @@ struct _EpgProviderInterface
 
   gboolean        (*add_code)   (EpgProvider  *self,
                                  const gchar  *code_str,
+                                 gint64       *time_added,
                                  GError      **error);
   gboolean        (*clear_code) (EpgProvider  *self,
                                  GError      **error);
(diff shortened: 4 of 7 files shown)


libeos-payg/manager.c
```diff
@@ -649,12 +650,12 @@ set_expiry_time (EpgManager *self,
...
  * @now will typically be the value of (g_get_real_time() / G_USEC_PER_SEC).
  * It is parameterised to allow easy testing. */
-static void
+static gint64
 extend_expiry_time (EpgManager *self,
                     guint64     now_secs,
                     EpcPeriod   period)
 {
-  guint64 span_secs;
+  gint64 span_secs;
```
wjt: Changing the type of this variable means the `span_secs = G_MAXUINT64` statement below is illegal, because `G_MAXUINT64` does not fit into a signed `gint64`.

libeos-payg/tests/manager.c
```diff
@@ -204,6 +204,7 @@ test_manager_disabled (Fixture *fixture,
     remove_path (fixture->key_path);
 
   g_autoptr(GError) error = NULL;
+  gint64* time_added = 0;
```
wjt:
```suggestion
  gint64 *time_added = 0;
```

and in every other function where this is added.

libeos-payg/manager-service.c
```diff
@@ -835,17 +835,18 @@ epg_manager_service_manager_add_code (EpgManagerService     *self,
...
 
   guint64 now_secs = g_get_real_time () / G_USEC_PER_SEC;
-  epg_provider_add_code (self->provider, code_str, now_secs, &local_error);
+  epg_provider_add_code (self->provider, code_str, now_secs, &time_added, &local_error);
 
   if (local_error != NULL)
     g_dbus_method_invocation_return_gerror (invocation, local_error);
   else
-    g_dbus_method_invocation_return_value (invocation, NULL);
+    g_dbus_method_invocation_return_value (invocation, g_variant_new("(x)", time_added));
```
wjt:
```suggestion
    g_dbus_method_invocation_return_value (invocation, g_variant_new ("(x)", time_added));
```

libeos-payg/provider.h
```diff
@@ -35,6 +35,7 @@ struct _EpgProviderInterface
   gboolean        (*add_code)   (EpgProvider  *self,
                                  const gchar  *code_str,
                                  guint64       now_secs,
+                                 gint64       *time_added,
```
wjt: We'd need a corresponding PR for `eos-payg-nonfree` to add this parameter to that provider.

libeos-payg/manager.c
```diff
@@ -814,7 +817,7 @@ epg_manager_add_code (EpgProvider   *provider,
   g_array_sort (self->used_codes, used_codes_sort_cb);
 
   /* Extend the expiry time. */
-  extend_expiry_time (self, now_secs, period);
+  *time_added = extend_expiry_time (self, now_secs, period);
```
wjt: There should be a `g_return_val_if_fail (time_added != NULL, FALSE)` at the top of this function.

libeos-payg/provider.c
```diff
@@ -172,7 +173,7 @@ epg_provider_add_code   (EpgProvider  *self,
 
   g_assert (iface->add_code != NULL);
 
-  return iface->add_code (self, code_str, now_secs, error);
+  return iface->add_code (self, code_str, now_secs, time_added, error);
```
wjt: This function should either `g_return_val_if_fail (time_added != NULL, FALSE)` or it should arrange for a dummy variable to be passed to `iface->add_code` if the parameter is `NULL`. The former is more straightforward. This parameter should also be added to the documentation comment above this function.

libeos-payg/tests/manager.c
```diff
@@ -213,7 +214,7 @@ test_manager_disabled (Fixture *fixture,
   g_assert_cmpuint (0, ==, epg_provider_get_expiry_time (fixture->provider));
   g_assert_cmpuint (0, ==, epg_provider_get_rate_limit_end_time (fixture->provider));
 
-  ret = epg_provider_add_code (fixture->provider, "00000000", 0, &error);
+  ret = epg_provider_add_code (fixture->provider, "00000000", 0, &time_added, &error);
```
wjt: All the tests which make assertions about how the expiry time changes after calling add_code should make the same assertion about the value returned in time_added.

libeos-payg/manager.c
mazen-asef: Ah, you are right - should I alter the type of the set_expiry_time function as well to keep it consistent across the code?

wjt: I think you should either change the type of `set_expiry_time()`'s parameter to `gint64` and add `g_return_if_fail (span_secs >= 0);`  to the top; or leave this variable as `guint64` and add `g_return_val_if_fail (span_secs >= G_MAXINT64, G_MAXINT64)` at the bottom of this function. Personally I think I'd prefer the first option. It's really awkward that the AddCode method needs to return a signed value, but it really does have to be because the provider in eos-payg-nonfree can reduce the amount of credit in response to a call to AddCode.

Either way, you'll need to change the EPC_PERIOD_INFINITE branch of the switch.

libeos-payg/tests/manager.c
```diff
@@ -606,6 +617,7 @@ test_manager_error_rate_limit (Fixture *fixture,
   manager_new (fixture);
 
   g_autoptr(GError) error = NULL;
+  gint64 *time_added = NULL;
```
wjt: Surely this (and the others in this file) should be `gint64 time_added = 0` not `gint64 *`?

mazen-asef: Ah, you are right, just read a bit about it and it does not seem necessary here.

mazen-asef:
Having merged with master brought a few other issues with the guint64 x gint64 problem. Isn't it bound to break if we making too many operations comparing both in the future, once it reaches the tipping point between the integers? Writing the tests and returns for the infinite code seems like false positives.
Or I shouldn't worry too much if it's not the maxint constant?

libeos-payg/manager-service.c
```diff
@@ -835,16 +835,17 @@ epg_manager_service_manager_add_code (EpgManagerService     *self,
                                       GDBusMethodInvocation *invocation)
 {
   g_autoptr(GError) local_error = NULL;
+  gint64 time_added = NULL;
```
wjt: `NULL` is not a `gint64`. You want `0`.

libeos-payg/tests/manager.c
```diff
@@ -612,18 +620,20 @@ test_manager_error_malformed (Fixture *fixture,
...
 
   /* Just use an obviously-wrong code. The nuances are tested in libeos-payg-codes. */
-  ret = epg_provider_add_code (fixture->provider, "abcdefgh", &error);
+  ret = epg_provider_add_code (fixture->provider, "abcdefgh", &time_added, &error);
   g_assert_error (error, EPG_MANAGER_ERROR, EPG_MANAGER_ERROR_INVALID_CODE);
   g_assert_false (ret);
 
   expiry_after_code = epg_provider_get_expiry_time (fixture->provider);
   g_assert_cmpint (expiry, ==, expiry_after_code);
+  g_assert_null (time_added);
```
wjt: `time_added` is not a pointer. I think you want `g_assert_cmpint (time_added, ==, 0)` (or no assertion).

wjt:
I think it is okay to replace references to G_MAXUNT64 with G_MAXINT64 in the test cases for adding an infinite code. G_MAXINT64 seconds after a computer boots is still an unimaginably large number.

libeos-payg/manager.c
```diff
@@ -645,7 +651,7 @@ set_expiry_time (EpgManager *self,
   guint64 base_secs = (now_secs < old_expiry_time_secs) ? old_expiry_time_secs : now_secs;
 
   /* Clamp to the end of time instead of overflowing. */
-  self->expiry_time_secs = (base_secs <= G_MAXUINT64 - span_secs) ? base_secs + span_secs : G_MAXUINT64;
+  self->expiry_time_secs = (base_secs <= G_MAXINT64 - span_secs) ? base_secs + span_secs : G_MAXUINT64;
```
wjt: `self->expiry_time_secs` is still a `guint64` so this change is not necessary.

libeos-payg/tests/manager.c
pwithnall: Is `-Wint-conversion` not enabled, or is someone ignoring warnings?

libeos-payg/manager-service.c
pwithnall: Is `-Wint-conversion` not enabled, or is someone ignoring warnings?

libeos-payg/tests/manager.c
mazen-asef: I'm not used to working with it, so it does seem that i've been summarily ignoring warnings...

mazen-asef: Oh, yeah, it's giving warnings because I end up comparing guint with gint - even though these two values are comparable it does break when we're assuming the user added a infinite code value.

libeos-payg/provider.h
wjt: https://github.com/endlessm/eos-payg-nonfree/pull/8

libeos-payg/tests/manager.c
```diff
@@ -612,18 +623,20 @@ test_manager_error_malformed (Fixture *fixture,
...
 
   /* Just use an obviously-wrong code. The nuances are tested in libeos-payg-codes. */
-  ret = epg_provider_add_code (fixture->provider, "abcdefgh", &error);
+  ret = epg_provider_add_code (fixture->provider, "abcdefgh", &time_added, &error);
   g_assert_error (error, EPG_MANAGER_ERROR, EPG_MANAGER_ERROR_INVALID_CODE);
   g_assert_false (ret);
 
   expiry_after_code = epg_provider_get_expiry_time (fixture->provider);
   g_assert_cmpint (expiry, ==, expiry_after_code);
+  g_assert_false (time_added);
```
wjt: It's not a boolean! 
```suggestion
  g_assert_cmpint (time_added, ==, 0);
```

libeos-payg/tests/manager.c
```diff
@@ -698,19 +713,20 @@ test_manager_error_rate_limit (Fixture *fixture,
...
       code_str = get_next_code (fixture);
-      ret = epg_provider_add_code (fixture->provider, code_str, &error);
+      ret = epg_provider_add_code (fixture->provider, code_str, &time_added, &error);
       g_assert_no_error (error);
       g_assert_true (ret);
     }
 
   expiry_after_code = epg_provider_get_expiry_time (fixture->provider);
   g_assert_cmpuint (expiry + (64 * 5), ==, expiry_after_code);
+  g_assert_cmpint (5, ==, time_added);
```
wjt: This should be inside the loop, otherwise you're only testing the first call.

libeos-payg/provider.c
```diff
@@ -155,34 +155,41 @@ epg_provider_default_init (EpgProviderInterface *iface)
...
+ * populated by the span of seconds added by @code_str. If @code_str fails
+ * verification or cannot be added, an error will be returned.
  *
  * Calls to this function may be rate limited: if too many attempts are made within
  * a given time period, %EPG_MANAGER_ERROR_TOO_MANY_ATTEMPTS will be returned
  * until that period expires. The rate limiting history is reset on a successful
- * verification of a code.
+ * verification of a code. 
+ *
+ * The reason @time_added is a signed int is due to the fact some providers may use
```
wjt:
```suggestion
 * @time_added is signed because some implementations can reduce the available credit when you add a new code.
```

libeos-payg/provider.c
```diff
@@ -155,34 +155,41 @@ epg_provider_default_init (EpgProviderInterface *iface)
...
+ * verification or cannot be added, an error will be returned.
  *
  * Calls to this function may be rate limited: if too many attempts are made within
  * a given time period, %EPG_MANAGER_ERROR_TOO_MANY_ATTEMPTS will be returned
  * until that period expires. The rate limiting history is reset on a successful
- * verification of a code.
+ * verification of a code. 
+ *
+ * @time_added is signed because some implementations can reduce the available credit when you add a new code.
+ * the add_code funcionality to actually remove time from the credit available.
```
treitter: It looks like this last line was accidentally included after re-writing the original sentence.

treitter:
I think this is ready once @wjt is happy with the latest changes.

wjt:
@mazen-asef 
> This branch has conflicts that must be resolved
Please rebase it.

PR #32, opened 2019-01-07
eos-payg-ctl: Understand CLOCK_BOOTTIME properties

mwleeds:
Now that the "ExpiryTime" and "RateLimitEndTime" properties of
EpgProvider are in terms of CLOCK_BOOTTIME instead of being UNIX
timestamps, modify eos-payg-ctl to interpret them correctly. For now
this is implemented using a helper program until we get Python 3.7 on
the stable branch, which will give us time.CLOCK_BOOTTIME (to be used
with time.clock_gettime()).

### DIFF
--- a/debian/eos-payg-data.install
+++ b/debian/eos-payg-data.install
@@ -1,2 +1,3 @@
 usr/share/locale
 usr/bin/eos-payg-ctl
+usr/bin/get-clock-boottime
--- a/eos-payg-ctl/eos-payg-ctl
+++ b/eos-payg-ctl/eos-payg-ctl
@@ -20,8 +20,10 @@
 import argparse
 import contextlib
 import datetime as dt
+import time
 import posix
 import sys
+import subprocess
 from dateutil import tz
 from gi.repository import GLib, Gio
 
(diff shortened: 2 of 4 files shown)


eos-payg-ctl/get-clock-boottime.c
```diff
@@ -0,0 +1,46 @@
...
+#include <unistd.h>
+#include <sys/timerfd.h>
+
+/* Print the current value of CLOCK_BOOTTIME in seconds */
+int main(void)
+{
+  struct timespec ts;
+  int result;
+
+  result = clock_gettime (CLOCK_BOOTTIME, &ts);
```
treitter: nitpick: with such a simple program, I would collapse the definition and assignment into a single line. Or, at least, delete the blank line that follows this

mwleeds:
Here's a follow-up task for switching to pure Python: https://phabricator.endlessm.com/T25065

eos-payg-ctl/eos-payg-ctl
```diff
@@ -33,6 +35,19 @@ ERROR_DOMAIN = "com.endlessm.Payg1.Error"
 DBUS_PROPERTIES_INTERFACE = "org.freedesktop.DBus.Properties"
 
 
+def __get_utc_from_boottime(timestamp):
+    #FIXME: Replace this with time.clock_gettime(time.CLOCK_BOOTTIME)
```
treitter: If you've already created the task to do this work, include its ID here

debian/eos-payg-data.install
```diff
@@ -1,2 +1,3 @@
 usr/share/locale
 usr/bin/eos-payg-ctl
+usr/bin/get-clock-boottime
```
pwithnall: Nitpick: Helper executables like this generally go in a subdirectory of `lib` or `libexec`, because they don’t need to appear in the `PATH`. i.e. `/usr/lib/eos-payg1/eos-payg-get-clock-boottime`. Maybe worth fixing if you have a moment, but not critical.

eos-payg-ctl/eos-payg-ctl
```diff
@@ -33,6 +35,20 @@ ERROR_DOMAIN = "com.endlessm.Payg1.Error"
...
+    # once we have Python 3.7
+    # https://phabricator.endlessm.com/T25065
+    try:
+        current_boottime = int(subprocess.check_output('get-clock-boottime'))
```
wjt: `time.clock_gettime(7)` would also have worked on Python 3.5

pwithnall: IMO, that would be preferable enough to warrant dropping `get-clock-boottime` immediately.

PR #33, opened 2019-01-09
eos-payg-ctl: Use pure Python to get CLOCK_BOOTTIME

mwleeds:
It turns out you can pass the value "7" directly to time.clock_gettime()
rather than waiting for Python 3.7 to define time.CLOCK_BOOTTIME as 7.
So do this rather than using a helper program written in C. This is also
needed because the C program was added to the eos-payg-data package
which is architecture independent, so an ARM binary ended up in the
Intel ostree (leading to a runtime error).


eos-payg-ctl/eos-payg-ctl
```diff
@@ -39,13 +38,7 @@ def __get_utc_from_boottime(timestamp):
...
     # once we have Python 3.7
     # https://phabricator.endlessm.com/T25065
-    try:
-        current_boottime = int(subprocess.check_output('get-clock-boottime'))
-    except subprocess.CalledProcessError as e:
-        print("Error calling get-clock-boottime: {}".format(e.output), file=sys.stderr, flush=True)
-        raise SystemExit(1)
-
-    timestamp_delta = timestamp - current_boottime
+    timestamp_delta = timestamp - time.clock_gettime(7)
```
wjt: My WIP patch used `getattr(time, 'CLOCK_BOOTTIME', 7)` but this works too.

wjt: It would also have been good to explain where '7' comes from.

PR #34, opened 2019-01-09
eos-payg-ctl: improve conversion from BOOTTIME to human-readable timestamp

wjt:
Some stylistic stuff, and one very minor improvement to use the same offset for the two properties, rather than introducing fractional differences. All absolutely inessential.

### DIFF
--- a/eos-payg-ctl/eos-payg-ctl
+++ b/eos-payg-ctl/eos-payg-ctl
@@ -40,15 +40,29 @@ DBUS_PROPERTIES_INTERFACE = "org.freedesktop.DBus.Properties"
 CLOCK_BOOTTIME = getattr(time, 'CLOCK_BOOTTIME', 7)
 
 
-def __format_boottime_timestamp(boottime_timestamp):
-    # Convert from CLOCK_BOOTTIME to seconds since the epoch
-    timestamp_delta = boottime_timestamp - time.clock_gettime(CLOCK_BOOTTIME)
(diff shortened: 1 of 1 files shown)


eos-payg-ctl/eos-payg-ctl
```diff
@@ -40,15 +40,29 @@ DBUS_PROPERTIES_INTERFACE = "org.freedesktop.DBus.Properties"
...
-def __format_boottime_timestamp(boottime_timestamp):
-    # Convert from CLOCK_BOOTTIME to seconds since the epoch
-    timestamp_delta = boottime_timestamp - time.clock_gettime(CLOCK_BOOTTIME)
-    utc_timestamp = time.time() + timestamp_delta
-
-    # Format that as a human-readable local time
-    expiry_utc = dt.datetime.fromtimestamp(utc_timestamp, tz=tz.tzutc())
-    expiry_local = expiry_utc.astimezone(tz.tzlocal())
-    return str(expiry_local)
+def __make_boottime_formatter():
```
mwleeds: Personally I think it would be more readable to have `now_boottime` and `now_utc` initialize statically rather than using a function to get a function, but it's not worth holding up the merge over this.

wjt: That's fair. I really like higher-order functions, and what I like about this approach is that `now_boottime` and `now_utc` don't leak out into global scope, but perhaps you're right that it makes it less readable for little benefit.

PR #35, opened 2019-01-16
eos-payg-ctl: use time.CLOCK_BOOTTIME directly

wjt:
This adds a dependency on Python 3.7. (Actually, it's possible that this
was added in Python 3.6, but that version was never in Endless OS.)

Please do not merge this until PAYG development on the eos3.5 branch has wound down.

https://phabricator.endlessm.com/T25065


mwleeds:
We have Python 3.7 on master; I believe this is safe to merge

PR #37, opened 2019-01-24
manager: Save state after migrating state files

mwleeds:
We want the EpgManager to save state to the disk upon startup, because
otherwise an unclean shutdown would leave us in the wrong state upon the
next boot. However the call to epg_manager_save_state_async() is in
file_load_cb() in an if block that only gets executed if the two state
files used to store the expiry time have been loaded, which means it
doesn't get executed if we're migrating from the old single state file
(migrating from expiry-time to clock-time and expiry-seconds).

So move the save_state_async() call to set_expiry_time() so that it
happens every time the expiry time is updated. This applies in the
migration case above but also a few other code paths, such as when
there's an error loading the state data. This commit also removes the
save_state_async() call in add_code(), because I think the only time the
state save is necessary is when the added code affected the expiry time.

https://phabricator.endlessm.com/T25138

### DIFF
--- a/libeos-payg/manager.c
+++ b/libeos-payg/manager.c
@@ -629,9 +629,16 @@ check_expired_cb (gpointer user_data)
 /* Set the #EpgManager:expiry-time to
  * `MIN (G_MAXUINT64, MAX (@now, #EpgManager:expiry-time) + @span)` and set
  * the #GSource expiry timer to the new expiry time. Everything is handled in
- * seconds. */
+ * seconds.
+ *
+ * @task, if provided, will have epg_multi_task_increment() called on it (and
+ * decremented in the callback of the state save). If you are going to kick off
+ * a state save soon after calling this, set @save_state to %FALSE to prevent
+ * redundant I/O. */
 static void
 set_expiry_time (EpgManager *self,
+                 GTask      *task,
+                 gboolean    save_state,
                  guint64     now_secs,
                  guint64     span_secs)
 {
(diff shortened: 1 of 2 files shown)


libeos-payg/manager.c
```diff
@@ -857,16 +872,6 @@ epg_manager_add_code (EpgProvider   *provider,
   /* Reset the rate limiting history, since the code was successful. */
   clear_rate_limiting (self);
 
-  /* Kick off an asynchronous save.
```
wjt: The commit message says:

> This commit also removes the
> save_state_async() call in add_code(), because I think the only time the
> state save is necessary is when the added code affected the expiry time.

The state also includes the list of used codes (fine, any new code being added will also affect the expiry time) and, I believe, the rate limit end time. Currently it looks like the latter doesn't cause the state to be flushed to disk, which is (I think) a bug, but it's an existing one since that's checked before this point.

mwleeds: The rate limit end time is not part of the state that's saved to disk, as far as I can tell. I'll look at adding back the save on `add_code()`, hopefully without doing multiple identical disk writes in a row.

wjt:
Is it straightforward to test the "state not saved on migration followed by unclean shutdown" case?

libeos-payg/manager.c
```diff
@@ -914,9 +919,19 @@ internal_save_state_cb (GObject      *source_object,
   EpgManager *self = EPG_MANAGER (source_object);
   EpgProvider *provider = EPG_PROVIDER (self);
   g_autoptr(GError) local_error = NULL;
+  g_autoptr(GTask) task = NULL;
+
+  if (user_data != NULL)
+    task = G_TASK (user_data);
```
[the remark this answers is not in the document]
pwithnall: Correct, `G_TASK()` boils down to `G_TYPE_CHECK_INSTANCE_CAST()`, which always returns `NULL` without a warning if it’s passed `NULL`.

mwleeds:
> Is it straightforward to test the "state not saved on migration followed by unclean shutdown" case?

I'm not sure how you'd prevent the shutdown code from running (except maybe an env var). There's not a separate process you can `kill -9`

PR #38, opened 2019-01-28
eos-payg-csv: bulk generate time codes for keys

treitter:
Given a specially-formatted .csv file (see README), read any number of
private keys and generate time codes for each key.

The generated time codes are output as one .csv file per key.

In case there are multiple keys for any given device ID, codes are only
generated for the last key.

https://phabricator.endlessm.com/T25205


eos-payg-csv/eos-payg-csv
```diff
@@ -0,0 +1,96 @@
+#!/usr/bin/env python3
```
pwithnall: Missing a copyright header.

eos-payg-csv/eos-payg-csv
```diff
@@ -0,0 +1,96 @@
...
+
+import csv
+import os
+import shutil
+import subprocess
+import sys
+
+PERIODS = "1d 2d 3d 4d 5d 30d 60d 90d 365d infinite".split()
```
pwithnall: Why not just write this as a list to begin with?
```
PERIODS = ['1d', '2d', '3d', …]
```

eos-payg-csv/eos-payg-csv
```diff
@@ -0,0 +1,96 @@
...
+                csv_writer.writerow(row)
+        files_written.append(csv_filename)
+
+    print("created CSV files:")
+    for f in files_written:
+        print(f)
+
+if __name__ == "__main__":
+    if len(sys.argv) != 2:
+        print("usage: {} ACCT_KEY_CSV_FILE".format(sys.argv[0]), file=sys.stderr)
```
pwithnall: If someone calls `eos-payg-csv --help` it’s going to output to a file called `--help`. Probably best to just add `argparse` support rather than inventing your own.

eos-payg-csv/eos-payg-csv
```diff
@@ -0,0 +1,96 @@
...
+        return "1 day"
+
+    return "{0} days".format(days)
+
+def key_to_tmp_filename(key):
+    return os.path.join(TMPDIR, "{}.key".format(key))
+
+def rm_and_mkdir(dirname):
+    if os.path.isdir(dirname) and not os.path.islink(dirname):
+        if len(os.listdir(dirname)) == 0:
```
pwithnall: Do you need this `listdir` check? Shouldn’t `shutil.rmtree(dirname)` do the right thing if called on an empty directory and delete it?

eos-payg-csv/eos-payg-csv
```diff
@@ -0,0 +1,96 @@
...
+            os.rmdir(dirname)
+        else:
+            shutil.rmtree(dirname)
+    os.mkdir(dirname)
```
pwithnall: In the case that `os.path.islink(dirname)` is true, this `mkdir()` will still be called, and will fail (I think). Is that desired?

Why are links handled specially in this function? Could do with a comment.

eos-payg-csv/eos-payg-csv
```diff
@@ -0,0 +1,96 @@
...
+            id_to_key[row[0]] = row[-1]
+
+    # strip out an entry for the CSV header, if it exists
+    id_to_key.pop("device_id", None)
+
+    try:
+        rm_and_mkdir(TMPDIR)
+        rm_and_mkdir(OUTDIR)
+    except OSError:
+        print ("Creation of the directory %s failed" % TMPDIR)
```
pwithnall: Nitpick: Python style requires no space before `(` in a function call. Maybe run `python3-pep8`, `python3-pyflakes` and `pylint-3` over this file.

eos-payg-csv/README
```diff
@@ -0,0 +1,29 @@
...
+
+header row
+----------
+device_id,code 1, code 2, code 3,key
+
+device row
+----------
+DEVICE_ID,CODE_1,CODE_2,CODE_3,"KEY"
+
+where:
```
pwithnall: Would be useful to provide a sample CSV file in the README, to illustrate the description.

eos-payg-csv/eos-payg-csv
```diff
@@ -0,0 +1,96 @@
...
+    with open(id_to_key_file, 'r') as f:
+        reader = csv.reader(f, quotechar='"')
+        for row in reader:
+            # we specifically want the last key for each device ID since there
+            # may be multiple entries for each device ID. Using a dict
+            # guarantees this.
+            id_to_key[row[0]] = row[-1]
+
+    # strip out an entry for the CSV header, if it exists
+    id_to_key.pop("device_id", None)
```
pwithnall: The README says the header must exist. This parser doesn’t seem to be doing any validation; is that deliberate?

eos-payg-csv/eos-payg-csv
```diff
@@ -0,0 +1,96 @@
...
+
+    try:
+        rm_and_mkdir(TMPDIR)
+        rm_and_mkdir(OUTDIR)
```
pwithnall: Thinking about it a bit more, why does this delete the directories at all? Normally utilities would error out if a directory they expect to ‘own’ already exists, to avoid potentially deleting user data if the utility is run in the wrong place.

eos-payg-csv/eos-payg-csv
```diff
@@ -0,0 +1,96 @@
...
+
+    for id in id_to_key:
+        key_filename = key_to_tmp_filename(id)
+        with open(key_filename, 'w') as f:
+            f.write(id_to_key[id])
+
+        period_to_codes = {}
+        for period in PERIODS:
+            output = subprocess.run(['eos-payg-generate', key_filename, period],
+                    stdout=subprocess.PIPE).stdout.decode('utf-8')
```
pwithnall: Nitpick: Indentation seems to be wonky here. `pep8` would probably complain.

eos-payg-csv/eos-payg-csv
```diff
@@ -0,0 +1,96 @@
...
+            codes = []
+            for code in output.split('\n'):
+                codes.append(code)
```
pwithnall: Aren’t these three lines entirely equivalent to `codes = output.split('\n')`?

eos-payg-csv/eos-payg-csv
```diff
@@ -0,0 +1,96 @@
...
+
+        zipped = zip(*period_to_codes.values())
+
+        csv_filename = os.path.join(OUTDIR, "{0}.csv".format(id))
+        with open(csv_filename, 'w', newline='') as csvfile:
+            csv_writer = csv.writer(csvfile)
+
+            headers = []
+            for period in PERIODS:
+                headers.append(period_to_display_str(period))
```
pwithnall: These three lines could be simplified as `headers = [period_to_display_str(period) for period in PERIODS]`

eos-payg-csv/eos-payg-csv
```diff
@@ -0,0 +1,96 @@
...
+            csv_writer.writerow(headers)
+
+            for row in list(zipped):
+                csv_writer.writerow(row)
+        files_written.append(csv_filename)
+
+    print("created CSV files:")
```
pwithnall:
```suggestion
    print('Created CSV files:')
```

eos-payg-csv/eos-payg-csv
pwithnall: Why are the directories needed at all?

And if you need a temporary directory, would [`TemporaryDirectory`](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory) be better?

eos-payg-csv/eos-payg-csv
wjt: You want `check=True` so this raises an exception if the subcommand fails.

eos-payg-csv/eos-payg-csv
wjt: Or `codes = output.splitlines()` :)

eos-payg-csv/eos-payg-csv
treitter: My original idea was that it would be tidier in case you're running uninstalled but, yeah, probably needlessly complicated so I've cut them out.

eos-payg-csv/eos-payg-csv
treitter: Fixed

eos-payg-csv/eos-payg-csv
treitter: I'll see your `splitlines()` and raise you a list comprehension (for reasons explained in a comment block above this new code):

```
codes = ["'{}".format(code) for code in output.splitlines()]
```

This is a case where I miss Java's overloaded + operator for simple string joins. Maybe Python has a shorter option than `format()`.

eos-payg-csv/eos-payg-csv
treitter: hah, I wrote the list comprehension just above this before getting to this comment :)

Fixed locally.

eos-payg-csv/eos-payg-csv
treitter: Fixed locally

eos-payg-csv/eos-payg-csv
treitter: Thanks, fixed locally

treitter:
Note that the current state of this branch has a few `TODO`s related to making the installed tests work. I'll address those before merging. But the test suite works when run manually, uninstalled.

eos-payg-csv/eos-payg-csv
```diff
@@ -0,0 +1,180 @@
...
+class CsvParseError(Exception):
+    pass
+
+def get_csv_rows(id_to_key_file):
+    rows = None
+    try:
+        with open(id_to_key_file, 'r') as f:
+            rows = list(csv.reader(f, quotechar='"'))
+    except OSError as ose:
+        raise CsvParseError("failed to read ACCT_KEY_CSV_FILE: {}".format(ose))
```
mwleeds: It would make more sense to put the actual file name in the error message, I think

eos-payg-csv/eos-payg-csv
```diff
@@ -0,0 +1,180 @@
...
+
+    rows = get_csv_rows(id_to_key_file)
+    for row in rows:
+        if first_row:
+            # normalize in case there are \r characters, etc.
+            header_norm = ",".join(row).strip()
+            header_expected = 'device_id,code1,code2,code3,key'
+            if header_norm != header_expected:
+                raise CsvParseError('''Provided CSV invalid:
+expected header:
```
mwleeds: nitpick: the lack of indentation of this string makes the code ugly. We could declare it statically at the top of the file?

eos-payg-csv/eos-payg-csv
```diff
@@ -0,0 +1,180 @@
...
+    {}
+got header:
+    {}'''.format(header_expected, row))
+
+            first_row = False
+        else:
+            expected_col_len = 5
```
mwleeds: It makes more sense to declare this and `expected_key_min` outside the for loop I think

eos-payg-csv/eos-payg-csv
```diff
@@ -0,0 +1,180 @@
...
+                csv_writer.writerow(row)
+        files_written.append(csv_filename)
+
+    return files_written
+
+def main(argv):
+    try:
+        id_to_key = acct_csv_get_id_to_key(argv[0])
+    except CsvParseError as cpe:
+        print("failed to read ACCT_KEY_CSV_FILE: {}".format(cpe),
```
mwleeds: As above, I think this should be the actual filename rather than ACCT_KEY_CSV_FILE

eos-payg-csv/eos-payg-csv
```diff
@@ -0,0 +1,180 @@
...
+
+    if days == "1":
+        return "1 day"
+
+    return "{0} days".format(days)
+
+def key_to_filename(key):
+    return "{}.key".format(key)
+
+def rm_and_mkdir(dirname):
```
mwleeds: This function isn't used at all anymore

eos-payg-csv/README.md
```diff
@@ -0,0 +1,43 @@
...
+============
+Usage:
+    eos-payg-csv ACCT_KEY_CSV_FILE
+
+Read per-device `DEVICE_ID` and Pay As You Go private keys from
+`ACCT_KEY_CSV_FILE` and generate time codes for several time periods per key.
+
+For each `DEVICE_ID`, the last corresponding key that appears in
+`ACCT_KEY_CSV_FILE` will be used to generate these unlock codes and the output
+will be `codes/DEVICE_ID.csv`. Earlier keys are ignored in case a given computer
```
mwleeds: This needs to be updated now that the `codes` directory isn't created

eos-payg-csv/tests/eos-payg-csv.py
```diff
@@ -0,0 +1,168 @@
...
+        if 'G_TEST_SRCDIR' in os.environ:
+            self.__eos_payg_csv = \
+                os.path.join(os.environ['G_TEST_SRCDIR'], '..', 'eos-payg-csv')
+            self.__datadir = os.environ['G_TEST_SRCDIR']
+        else:
+            self.__eos_payg_csv = os.path.join('/', 'usr', 'bin',
+                                               'eos-payg-csv')
+            self.__datadir = os.path.join('/', 'usr', 'share', 'installed-tests',
+                                          'eos-payg-csv')
+        print('eos_payg_csv:', self.__eos_payg_csv)
```
mwleeds: I guess these should be hyphens rather than underscores

eos-payg-csv/tests/eos-payg-csv.py
```diff
@@ -0,0 +1,168 @@
...
+
+    def test_csv_wrong_columns(self):
+        """Test error handling when providing incorrectly-formatted file."""
+        info = self.runCsv(os.path.join(self.__datadir, 'wrong-columns.csv'))
+        out = info.stdout.decode('utf-8').strip()
+        self.assertIn('Provided CSV invalid:\nexpected header', out)
+        self.assertEqual(info.returncode, 1)  # EXIT_INVALID_OPTIONS
+
+    def test_csv_short_key(self):
+        """Test error handling when providing a CVS with a short key entry."""
```
mwleeds: typo: CSV

eos-payg-csv/tests/eos-payg-csv.py
```diff
@@ -0,0 +1,168 @@
...
+        info = self.runCsv(os.path.join(self.__datadir, 'short-key.csv'))
+        out = info.stdout.decode('utf-8').strip()
+        self.assertIn('Provided CSV invalid: minimum key length 64 bytes but '
+                      'CSV contains a key of 12 bytes', out)
+        self.assertEqual(info.returncode, 1)  # EXIT_INVALID_OPTIONS
+
+    def test_csv_valid(self):
+        """Test proper operation of the script for a 1-key input file."""
```
mwleeds: It would be nice to test a file with more than one key, and even more than one for the same device ID, but don't consider that a blocker for merging this unless you want to

pwithnall:
> @pwithnall how would you recommend I resolve that issue? I don't want to have to special-case the name of the program (within `eos-payg-csv`) depending on whether it was installed in a test environment.

Unconditionally call `eos-payg-generate-1` instead of `eos-payg-generate`. It’s always installed as the former in every environment; the latter is a symlink. It’s the same as the GStreamer command-line utilities.

eos-payg-csv/eos-payg-csv
pwithnall: This doesn’t seem to have been addressed.

eos-payg-csv/eos-payg-csv
pwithnall: You want [`dedent`](https://docs.python.org/3.1/library/textwrap.html#textwrap.dedent).

eos-payg-csv/eos-payg-csv
```diff
@@ -0,0 +1,180 @@
...
+        print(f)
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(
+        description='generate time codes for private keys in bulk',
+        formatter_class=argparse.RawTextHelpFormatter)
+    parser.add_argument('acct_key_csv_file', metavar='ACCT_KEY_CSV_FILE',
+                        type=str, nargs=1, help='''a file formatted as:
+
+device_id,code1,code2,code3,key
```
pwithnall: Could use [`dedent`](https://docs.python.org/3.1/library/textwrap.html#textwrap.dedent) here too.

eos-payg-csv/tests/eos-payg-csv.py
```diff
@@ -0,0 +1,168 @@
+#!/usr/bin/python3
+# -*- coding: utf-8 -*-
+#
+# Copyright © 2018 Endless Mobile, Inc.
```
pwithnall: Nitpick: It’s 2019!

eos-payg-csv/eos-payg-csv
treitter: fixed locally

eos-payg-csv/eos-payg-csv
treitter: fixed locally

eos-payg-csv/eos-payg-csv
treitter: Fixed locally

eos-payg-csv/eos-payg-csv
treitter: Fixed locally

eos-payg-csv/README.md
treitter: Fixed locally

eos-payg-csv/tests/eos-payg-csv.py
treitter: Fixed locally

eos-payg-csv/tests/eos-payg-csv.py
treitter: Fixed locally

eos-payg-csv/tests/eos-payg-csv.py
treitter: Fixed locally.

eos-payg-csv/eos-payg-csv
treitter: Fixed locally

eos-payg-csv/tests/eos-payg-csv.py
treitter: Fixed!

treitter:
@pwithnall @mwleeds ready for final review. The fixups don't have meaningful divisions (sorry), so you're best off just reading the net changes from this branch.

eos-payg-csv/eos-payg-csv
```diff
@@ -0,0 +1,183 @@
...
+            headers = [period_to_display_str(period) for period in PERIODS]
+            csv_writer.writerow(headers)
+
+            for row in list(zipped):
+                csv_writer.writerow(row)
+        files_written.append(csv_filename)
+
+    return files_written
+
+def main(key_files, eos_payg_generate):
```
mwleeds: Why take a list and then only use the first element? Why is `args.acct_key_csv_file` a list?

treitter: Any argument with `nargs` gets the arguments as a list (even if `nargs=1`).

But it turns out if you just omit `nargs` it will also require 1 argument but return it as a string, so I've fixed that.

PR #39, opened 2019-02-09
eos-payg-csv: ensure test files available for OBS

treitter:
Currently, the OBS builds are broken (though Jenkins builds are not)
because OBS performs the installed tests and does not see the required
data files.

This ensures the test files are available in that case.

https://phabricator.endlessm.com/T25205

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -49,7 +49,7 @@ Depends:
  ${misc:Depends},
  ${shlibs:Depends},
 Description: Pay As You Go Code Generator
- This package contains a utility to generate pay as you go codes.
+ This package contains utilities to generate pay as you go codes.
 
 Package: eos-payg-generate-tests
 Section: misc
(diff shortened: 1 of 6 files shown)


wjt:
I didn't realise Meson had that behaviour (or rather, at one point I did, but I'd forgotten).

While the bulk of these changes may not be **necessary**, is it not true that the test depends on the CSV files?

wjt:
Ah, Meson says "no, it's not correct":

```
eos-payg-csv/tests/meson.build:50:2: ERROR:  Depends items must be build targets.
```

treitter:
Once more, with less effort!

Jenkins test passed - @wjt could you please retest with OBS?

wjt:
No more easily than you can :)

treitter:
@wjt thanks for the suggestion. The build was breaking because of `dh_missing` but I created Debian packages so the installed tests could be created.

I don't want `eos-payg-csv` to be installed by default on user systems. That's the default - we have to edit something to include that package in the image and OSTree builds, right?

@wjt / @pwithnall could you please do one last brief look over the new commit I added? It's based closely on a commit Philip made in December for `eos-updater`.

pwithnall:
Why not put it in the existing `eos-payg-generate` package (and accompanying tests package)?

treitter:
> Why not put it in the existing `eos-payg-generate` package (and accompanying tests package)? They’re pretty highly related.

@pwithnall 

Fixed with the latest fixup and it passes on OBS.

OK to squash and merge?

wjt:
I squashed, and added one additional patch which I needed to build locally.

PR #41, opened 2019-03-06
T24501 Detect clock jumps

mwleeds:
https://phabricator.endlessm.com/T24501

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,3 @@
 /eos-payg-csv/tests/3F0A1564.csv
+tags
+__pycache__
--- a/debian/rules
+++ b/debian/rules
@@ -9,6 +9,7 @@ override_dh_auto_configure:
 		-- \
 		-Dinstalled_tests=true \
 		-Ddefault_library=static \
+		-Dtests=unsafe
 		$(NULL)
 
 # Remove LD_PRELOAD so we don't run with fakeroot. libgsystemservice's tests
--- a/libeos-payg/clock-jump-source.c
+++ b/libeos-payg/clock-jump-source.c
@@ -1,0 +1,156 @@
+/* -*- mode: C; c-file-style: "gnu"; indent-tabs-mode: nil; -*-
+ *
+ * Copyright © 2019 Endless Mobile, Inc.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at https://mozilla.org/MPL/2.0/.
+ *
+ * Alternatively, the contents of this file may be used under the terms of the
+ * GNU Lesser General Public License Version 2.1 or later (the "LGPL"), in
+ * which case the provisions of the LGPL are applicable instead of those above.
+ * If you wish to allow use of your version of this file only under the terms
+ * of the LGPL, and not to allow others to use your version of this file under
+ * the terms of the MPL, indicate your decision by deleting the provisions
+ * above and replace them with the notice and other provisions required by the
+ * LGPL. If you do not delete the provisions above, a recipient may use your
+ * version of this file under the terms of either the MPL or the LGPL.
+ */
+
+#include "config.h"
+
+#include <errno.h>
+#include <inttypes.h>
+#include <sys/timerfd.h>
+#include <unistd.h>
+#include <limits.h>
+#include <glib.h>
+#include <glib/gstdio.h>
+#include <gio/gio.h>
+
+#include <libeos-payg/clock-jump-source.h>
+
+/* This file is based on boottime-source.c */
+
+typedef struct {
+  GSource parent;
+
+  int fd;
+  gpointer tag;
+} EpgClockJumpSource;
+
+static gboolean
+epg_clock_jump_source_check (GSource *source)
+{
+  EpgClockJumpSource *self = (EpgClockJumpSource *) source;
+
+  return g_source_query_unix_fd (source, self->tag) != 0;
+}
+
+static gboolean
+epg_clock_jump_source_dispatch (GSource    *source,
+                                GSourceFunc callback,
+                                gpointer    user_data)
+{
+  EpgClockJumpSource *self = (EpgClockJumpSource *) source;
+  uint64_t n_expirations = 0;
+
+  if (callback == NULL)
+    {
+      g_warning ("ClockJump source dispatched without callback. "
+                 "You must call g_source_set_callback().");
+      return G_SOURCE_REMOVE;
+    }
+
+  /* Must read from the FD to reset its ready state. */
+  if (read (self->fd, &n_expirations, sizeof (n_expirations)) == 0 || errno != ECANCELED)
+    {
+      g_warning ("read() unexpectedly did not return ECANCELED for cancel-on-set timerfd: %s",
+                 errno == 0 ? "No error" : g_strerror (errno));
+      return G_SOURCE_REMOVE;
+    }
+
+  return callback (user_data);
+}
+
+static void
+epg_clock_jump_source_finalize (GSource *source)
+{
+  EpgClockJumpSource *self = (EpgClockJumpSource *) source;
+  g_autoptr(GError) local_error = NULL;
+
+  if (!g_close (self->fd, &local_error))
+    g_warning ("Failed to close timerfd: %s", local_error->message);
+  self->fd = -1;
+}
+
+static const GSourceFuncs epg_clock_jump_source_funcs = {
+  .check = epg_clock_jump_source_check,
+  .dispatch = epg_clock_jump_source_dispatch,
+  .finalize = epg_clock_jump_source_finalize,
+};
+
+/**
+ * epg_clock_jump_source_new:
+ * @error: return location for a #GError, or %NULL
+ *
+ * This creates a #GSource that will dispatch when the system clock jumps
+ * (changes discontinously) which can happen for example when the user sets it
+ * or NTP adjusts it. This is accomplished using the TFD_TIMER_CANCEL_ON_SET
+ * flag supported by timerfd_settime().
+ *
+ * There is an unlikely race here: if the clock jumps while this function is
+ * executing but before the timerfd is set up, that jump will be missed.
+ *
+ * @error will be set to a #GIOError if, for example, the process runs out
+ * of file descriptors.
+ *
+ * Returns: (transfer full): a new #GSource, or %NULL with @error set
+ * Since: 0.2.2
+ */
+GSource *
+epg_clock_jump_source_new (GError **error)
+{
+  g_autoptr(GSource) source = NULL;
+  EpgClockJumpSource *self = NULL;
+  int fd;
+  /* Unfortunately this will break in the year 2038 on platforms that define
+   * time_t to be 4 bytes. Hopefully we're not still supporting e.g. the EC-100
+   * at that point. */
+  struct itimerspec its = {
+    .it_interval = {
+      .tv_sec = TIME_MAX,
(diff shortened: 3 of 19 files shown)


libeos-payg/tests/clock-jump-source.c
```diff
@@ -0,0 +1,151 @@
...
+
+static gboolean
+timeout_cb (gpointer data)
+{
+  guint *called = (guint *) data;
+  *called += 1;
+  return G_SOURCE_CONTINUE;
+}
+
+/* Test the timeout firing once */
```
wjt: This test doesn't test a timeout firing, and it tests the source calling its callback twice, not once. I think given this tests both directions, and the same source firing more than once, the `many` test is redundant.

libeos-payg/tests/clock-jump-source.c
```diff
@@ -0,0 +1,151 @@
...
+  ret = clock_settime (CLOCK_REALTIME, &jump_ts);
+  g_assert_cmpint (ret, ==, 0);
+
+  while (called != 2)
+    g_main_context_iteration (NULL, TRUE);
+
+  g_source_destroy (source);
+  g_assert_cmpuint (called, ==, 2);
+
+  /* Undo our changes to the clock */
```
wjt: I don't think this is necessary. The test already does this, because `(- 10) + (+ 10) == 0`.

libeos-payg/tests/clock-jump-source.c
```diff
@@ -0,0 +1,151 @@
...
+test_once (void)
+{
+  g_autoptr(GError) local_error = NULL;
+  g_autoptr(GSource) source = NULL;
+  guint called = 0;
+  struct timespec original_ts;
+  struct timespec jump_ts;
+  int ret;
+
+  if (geteuid () != 0)
```
wjt: I wonder if this test should also be marked "slow" or something else to prevent it being run by default – I think I would be surprised if building a package as root caused my clock to be tinkered with.

libeos-payg/service.c
```diff
@@ -254,25 +262,77 @@ manager_new_cb (GObject      *source_object,
...
+  if (G_UNLIKELY (clock_gettime (CLOCK_REALTIME, &ts) != 0))
+    g_error ("clock_gettime() failed for CLOCK_REALTIME: %s", g_strerror (errno));
+  clock_realtime_secs_v1 = ts.tv_sec;
+  if (G_UNLIKELY (clock_gettime (CLOCK_BOOTTIME, &ts) != 0))
+    g_error ("clock_gettime() failed for CLOCK_BOOTTIME: %s", g_strerror (errno));
+  clock_boottime_secs_v1 = ts.tv_sec;
+
+  clock_jump_delta = ((clock_realtime_secs_v1 - self->clock_realtime_secs_v0) -
+                      (clock_boottime_secs_v1 - self->clock_boottime_secs_v0));
+  if (clock_jump_delta > 0)
```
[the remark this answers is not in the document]
wjt: I think you may need to allow a certain margin of error, because the pair of v1 and v0 measurements are not taken at exactly the same instant, so it's not impossible that there will be a spurious delta of 1 second (or even more, if the process happens to be descheduled at an inopportune moment.

libeos-payg/service.c
```diff
@@ -254,25 +262,77 @@ manager_new_cb (GObject      *source_object,
...
+clock_jump_cb (gpointer user_data)
+{
+  EpgService *self = EPG_SERVICE (user_data);
+  struct timespec ts;
+  gint64 clock_realtime_secs_v1;
+  gint64 clock_boottime_secs_v1;
+  gint64 clock_jump_delta;
+
+  if (G_UNLIKELY (clock_gettime (CLOCK_REALTIME, &ts) != 0))
+    g_error ("clock_gettime() failed for CLOCK_REALTIME: %s", g_strerror (errno));
```
wjt: This code is copy-pasted 4 times in this file. Perhaps consider a wrapper function which calls `clock_gettime`, `g_error` if it fails, and otherwise returns `tv_sec`?

libeos-payg/service.c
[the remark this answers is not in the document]
mwleeds: > What if it's negative, as it will be if CLOCK_REALTIME is set backwards in time?

You're right, it should be `if (clock_jump_delta > 0 || clock_jump_delta < 0)`

> I think you may need to allow a certain margin of error, because the pair of v1 and v0 measurements are not taken at exactly the same instant, so it's not impossible that there will be a spurious delta of 1 second (or even more, if the process happens to be descheduled at an inopportune moment.

I agree the measurement is imprecise but I don't see why we would get spurious deltas because this callback is only called if the kernel notices a discontinuous change to CLOCK_REALTIME, which I think would be unlikely to be less than a second.

.gitignore
```diff
@@ -1 +1,6 @@
 /eos-payg-csv/tests/3F0A1564.csv
+/tags
```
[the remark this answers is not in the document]
mwleeds: Made a fixup, and thanks for the tip

libeos-payg/service.c
mwleeds: You're right, that's much cleaner

libeos-payg/clock-jump-source.c
```diff
@@ -0,0 +1,156 @@
...
+{
+  g_autoptr(GSource) source = NULL;
+  EpgClockJumpSource *self = NULL;
+  int fd;
+  /* Unfortunately this will break in the year 2038 on platforms that define
+   * time_t to be 4 bytes. Hopefully we're not still supporting e.g. the EC-100
+   * at that point. */
+  struct itimerspec its = {
+    .it_interval = {
+      .tv_sec = sizeof(time_t) == 4 ? INT_MAX : LONG_MAX,
```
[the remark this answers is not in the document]
mwleeds: I added that compile time check.

I also read the lwn article you linked but left the y2038 problem. I looked around a bit but couldn't figure out how to make the `timerfd_settime` call safer. I really doubt Minis will still be working that far in the future (and I don't think we're using them for payg anyway) but if you feel strongly I can do further investigation.

libeos-payg/tests/clock-jump-source.c
mwleeds: Hmm, apparently systemd has a concept of an "unsafe" test but it's not a feature built into meson

pwithnall: You can [tag tests with zero or more ‘suites’ in Meson](https://mesonbuild.com/Unit-tests.html), and then pass `--suite` or `--no-suite` to `meson test`. If you grep the GLib source for `suite`, for example, you’ll see how we use it. (Some tests are tagged as `slow` or `flaky`: https://gitlab.gnome.org/GNOME/glib/blob/master/gio/tests/meson.build#L76.)

libeos-payg/clock-jump-source.h
```diff
@@ -0,0 +1,36 @@
...
+#pragma once
+
+#include <glib.h>
+
+G_BEGIN_DECLS
+
+#if SIZEOF_TIME_T == 8
+#  define TIME_MAX G_MAXINT64
+#elif SIZEOF_TIME_T == 4
+#  define TIME_MAX G_MAXINT32
```
pwithnall: Maybe add a `#warning Code is not Y2038 compliant` here?

libeos-payg/clock-jump-source.c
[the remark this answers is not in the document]
pwithnall: If you don’t think there are any platforms we ship which are a) not Y2038-safe in libc and the kernel; and b) running PAYG, then I’d suggest making it a compile error to try building eos-payg with a 4-byte `time_t`.

Failing that, file a task entitled “panic massively about Y2038 for any machines we still have to support with 4-byte time_t” and set its deadline to be 2036-01-01. (I’m only half joking.)

libeos-payg/clock-jump-source.h
pwithnall: Or `#error`, as per my other Y2038 comment

libeos-payg/service.c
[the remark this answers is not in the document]
wjt: Good point. Spurious no, but the delta might be slightly wrong. I don't see a way around this and it's unlikely to be significant unless something has gone dramatically wrong with the scheduler.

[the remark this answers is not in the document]
wjt: OTOH: if we're saying “this callback won't fire spuriously” then we don't need to check whether `delta` is non-zero. :-)

libeos-payg/clock-jump-source.h
wjt: This can't be `#error` because we do compile and ship `eos-payg` on ARM.

I don't really see the advantage of a `#warning` either because there's likely to be no action we can take for *years*, and no-one checks compile warnings on ARM anyway.

libeos-payg/clock-jump-source.c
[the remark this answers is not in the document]
wjt: When glibc supports 64-bit `time_t` on 32-bit platforms, then we'll have to take some action (for every Endless-specific project) anyway to rebuild them with some new `#define` to opt-in. (And then we'll discover that Mini uses a prehistoric vendor kernel that doesn't support the new syscalls anyway, but I'm sure it'll be EOL by 2038.)

I don't think we should do anything.

libeos-payg/tests/clock-jump-source.c
wjt: I think it'd be nice to do this but not a blocker.

I had a look at whether [libfaketime](https://github.com/wolfcw/libfaketime) supports timerfds (it doesn't) and whether time namespaces exist ([they don't, yet](https://lwn.net/Articles/766089/)). Boo.

mwleeds: I added it to an "unsafe" suite, but it still runs by default. Perhaps that's good enough to give people a hint

pwithnall: Why still run it by default?

mwleeds: (sorry for the delay; didn't see this comment)

I added a commit to disable unsafe tests unless `-Dtests=unsafe` is used. I suppose it's fine that most people won't run it since autopkgtest will.

wjt: > I suppose it's fine that most people won't run it since autopkgtest will.

[Nothing automatically runs these tests.](https://phabricator.endlessm.com/T18881)

wjt: (I agree with not running these unsafe tests by default.)

pwithnall: > Nothing automatically runs these tests.

(yet)

(patches welcome)

debian/rules
```diff
@@ -9,6 +9,7 @@ override_dh_auto_configure:
 		-- \
 		-Dinstalled_tests=true \
 		-Ddefault_library=static \
+		-Dtests=unsafe
```
wjt: This will be used on OBS so will run the unsafe tests on the OBS worker – I thought we wanted to avoid that?

wjt: Oh, never mind me. Provided OBS doesn't build packages as root, which I'm reasonably sure it doesn't, they won't run because `-rfakeroot` is removed below so the `geteuid () == 0` check will fail.

mwleeds: I was under the impression we wanted the unsafe test to be run by OBS during the autopkgtest step. Otherwise there's no point in adding the autopkgtest and the test would only be run manually by a determined developer (that is, almost never).

wjt: As mentioned [before](https://github.com/endlessm/eos-payg/pull/41#discussion_r266817175), nothing runs autopkgtests at the moment.

PR #42, opened 2019-03-29
service: Fix a seg fault on shutdown

mwleeds:
I tried to use g_clear_object() on a GSource pointer, but those are not
objects. Use g_source_destroy() followed by g_source_unref() instead to
avoid a seg fault on shutdown.

https://phabricator.endlessm.com/T24501

### DIFF
--- a/libeos-payg/service.c
+++ b/libeos-payg/service.c
@@ -126,7 +126,14 @@ epg_service_dispose (GObject *object)
 
   g_clear_object (&self->manager_service);
   g_clear_object (&self->provider);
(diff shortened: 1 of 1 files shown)


mwleeds:
You're correct there's a unit test that calls `epg_service_new` and garbage collects the object. I'm not sure we want to move the `epg_clock_jump_source_new` earlier in the object construction process because its callback operates on `self->provider` which isn't initialized until `epg_service_startup_complete`.

But certainly it would be great if the unit test would get the `EpgService` running on D-Bus and test more of the production code paths.

PR #43, opened 2019-05-16
manual-provision-payg: Fix path in sha256 file

mwleeds:
This was pointed out in the review but missed.

https://phabricator.endlessm.com/T23967


mwleeds:
Never mind, looks like we'd have to make build system changes and I don't see how to make the custom target that runs `sha256sum` run in the subdirectory. Also, the `Endless_Factory_Test.tar.sha256` in `eos-payg-nonfree` has the full path too.

wjt:
> Never mind, looks like we'd have to make build system changes and I don't see how to make the custom target that runs `sha256sum` run in the subdirectory. Also, the `Endless_Factory_Test.tar.sha256` in `eos-payg-nonfree` has the full path too.

I was **sure** I had arranged for this not to be the case but evidently not. Sorry for the noise on #30.

PR #44, opened 2019-06-18
eos-payg: Add a prefix and suffix property of the codes

mazen-asef:
Add a property that returns the prefix and sufix of the codes used,
making the interface of code insertion for user-friendly.

https://phabricator.endlessm.com/T25987

### DIFF
--- a/libeos-payg/manager-interface.h
+++ b/libeos-payg/manager-interface.h
@@ -131,12 +131,32 @@ static const GDBusPropertyInfo manager_interface_code_format =
   NULL,  /* annotations */
 };
 
+static const GDBusPropertyInfo manager_interface_code_format_prefix =
+{
+  -1,  /* ref count */
+  (gchar *) "CodeFormatPrefix",
+  (gchar *) "s",
+  G_DBUS_PROPERTY_INFO_FLAGS_READABLE,
+  NULL,  /* annotations */
+};
+
+static const GDBusPropertyInfo manager_interface_code_format_suffix =
+{
+  -1,  /* ref count */
+  (gchar *) "CodeFormatSuffix",
+  (gchar *) "s",
+  G_DBUS_PROPERTY_INFO_FLAGS_READABLE,
+  NULL,  /* annotations */
+};
+
 static const GDBusPropertyInfo *manager_interface_properties[] =
 {
   &manager_interface_expiry_time,
   &manager_interface_enabled,
   &manager_interface_rate_limit_end_time,
   &manager_interface_code_format,
+  &manager_interface_code_format_prefix,
+  &manager_interface_code_format_suffix,
   NULL,
 };
 
--- a/libeos-payg/provider.c
+++ b/libeos-payg/provider.c
@@ -121,6 +121,40 @@ epg_provider_default_init (EpgProviderInterface *iface)
   g_object_interface_install_property (iface, pspec);
 
   /**
+   * EpgProvider:code-format-prefix:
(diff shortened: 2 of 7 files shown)


libeos-payg/manager-interface.h
```diff
@@ -131,12 +131,32 @@ static const GDBusPropertyInfo manager_interface_code_format =
...
+  (gchar *) "CodeFormatPrefix",
+  (gchar *) "s",
+  G_DBUS_PROPERTY_INFO_FLAGS_READABLE,
+  NULL,  /* annotations */
+};
+
+static const GDBusPropertyInfo manager_interface_code_format_sufix =
+{
+  -1,  /* ref count */
+  (gchar *) "CodeFormatSufix",
```
wjt:
```suggestion
  (gchar *) "CodeFormatSuffix",
```
(and everywhere else "sufix" is used)

libeos-payg/provider.c
```diff
@@ -387,6 +421,48 @@ epg_provider_get_code_format (EpgProvider *self)
...
 }
 
+/**
+ * epg_provider_get_code_format_prefix:
+ * @self: a #EpgProvider
+ *
+ * Get the value of #EpgProvider:code-format-prefix
+ *
+ * Returns: the prefix of the code-format expected by this provider
+ * Since: 0.2.0
```
wjt: Not true.

libeos-payg/provider.c
```diff
@@ -387,6 +421,48 @@ epg_provider_get_code_format (EpgProvider *self)
...
+ */
+const gchar *
+epg_provider_get_code_format_prefix (EpgProvider *self)
+{
+  g_return_val_if_fail (EPG_IS_PROVIDER (self), NULL);
+
+  EpgProviderInterface *iface = EPG_PROVIDER_GET_IFACE (self);
+
+  g_assert (iface->code_format_prefix != NULL);
```
wjt: I'd rather have this function return `""` when the field is `NULL`. This would have the added upside that the existing version of eos-payg-nonfree would continue to work (just with these fields empty) rather than crashing.

libeos-payg/tests/manager.c
```diff
@@ -274,6 +274,36 @@ get_code_format (Fixture *fixture)
...
+ */
+static void
+test_manager_code_format_prefix (Fixture *fixture,
+                         gconstpointer data)
+{
+  manager_new (fixture);
+
+  const gchar *code_format_prefix = epg_provider_get_code_format_prefix (fixture->provider);
+
+  g_assert_nonnull (code_format_prefix);
```
wjt: You can make a stronger assertion here:

```suggestion
  g_assert_cmpstr (code_format_prefix, ==, "");
```

libeos-payg/provider.c
```diff
@@ -437,7 +437,7 @@ epg_provider_get_code_format_prefix (EpgProvider *self)
 
   EpgProviderInterface *iface = EPG_PROVIDER_GET_IFACE (self);
 
-  g_return_val_if_fail (iface->code_format_prefix == NULL, "");
+  g_return_val_if_fail (iface->code_format_prefix != NULL, "");
```
wjt: No, this just shouldn't be an assertion.

```c
if (iface->code_format_prefix != NULL)
  return iface->code_format_prefix;

return "";
```

mazen-asef: Oh, got it

PR #47, opened 2019-08-29
Error out if PAYG is not enabled when it should be

mwleeds:
Currently in epg_service_secure_init_sync() we have "g_assert (provider
!= NULL)" which means that eos-paygd hits an assertion failure in case
it's running on a non-PAYG system, or in case it's running on a PAYG
system using the filesystem to store state. In the latter case the
provider is only found to be enabled after the root pivot, and the
secure init happens before the root pivot.

This commit removes the assertion, and instead checks if no providers
have been found after the root pivot. If none were found and the
EOSPAYG_active EFI variable is set, we error out and force a poweroff
after 10 minutes. Otherwise the user could disable PAYG by zeroing out
the state data. The EFI variable check ensures that this security
measure only applies to Phase 4 systems, which will set EOSPAYG_active
during PAYG provisioning. On Phase 1 or 2 systems, we try falling back
to Endless PAYG and failing that eos-paygd gracefully exits.

https://phabricator.endlessm.com/T27581

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -14,6 +14,7 @@ Build-Depends:
  meson,
  python3-dbusmock,
  dracut,
+ libefivar-dev,
 
 Package: eos-paygd
 Section: misc
--- a/eos-paygd/main.c
+++ b/eos-paygd/main.c
@@ -198,17 +198,27 @@ main (int   argc,
 
   if (error != NULL)
     {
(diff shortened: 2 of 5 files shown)


mwleeds:
(force-pushed a typo fix)

PR #51, opened 2019-09-03
meson.build: Change warning_level to 3 and werror to true

mwleeds:
Increase the warning level to 3. Meson decides which -W cflags that
translates to. And set werror to true, which translates to -Werror.

The impetus for this is that any bugs in eos-paygd have potentially
severe impact (e.g. not enforce PAYG or crash and temporarily brick a
computer).

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -61,6 +61,9 @@ test_args = [
   '-Wcast-align',
   '-Wclobbered',
   '-Wno-declaration-after-statement',
+  # FIXME: Remove this -Wno-error after updating libpeas to 1.24 (libpeas
+  # header files mention GParameter)
+  '-Wno-error=deprecated-declarations',
   '-Wduplicated-branches',
   '-Wduplicated-cond',
   '-Wempty-body',
--- a/meson.build
+++ b/meson.build
@@ -7,7 +7,8 @@ project('eos-payg','c',
     # static.
     'default_library=static',
     'c_std=gnu11',
-    'warning_level=2',
+    'warning_level=3',
+    'werror=true',
   ],
 )
 
--- a/meson.build
+++ b/meson.build
@@ -113,6 +113,9 @@ test_args = [
 cc = meson.get_compiler('c')
 add_project_arguments(cc.get_supported_arguments(test_args), language: 'c')
 
+# Apply -Wno-pedantic to subprojects too
+add_global_arguments(cc.get_supported_arguments(['-Wno-pedantic']), language: 'c')
+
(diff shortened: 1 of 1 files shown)


mwleeds:
I don't think this is worth creating a ticket for.

pwithnall:
LGTM. [Jenkins failed](https://ci.endlessm-sf.com/job/obs-eos-payg-pr/233/console) due to some `-Wpedantic` warnings in libgsystemservice. I’ve [got a fix for them here](https://gitlab.gnome.org/pwithnall/libgsystemservice/merge_requests/16) (I think).

pwithnall:
The libgsystemservice fix has been merged upstream, so this PR needs updating to update the submodule commit. Then I think it should pass.

mwleeds:
Looks like the warning fixes in [this commit](https://gitlab.gnome.org/pwithnall/libgsystemservice/commit/c46252f0b097d824010d8a5ffbd3145217eb1ab9) still result in `-Wpedantic` warnings. I wonder if we should just add `-Wno-pedantic`?

Looks like we also need https://gitlab.gnome.org/pwithnall/libgsystemservice/merge_requests/17

pwithnall:
>  I wonder if we should just add `-Wno-pedantic`?

Generally I would go with a list of known-useful warning flags, rather than `-Wpedantic`, since `-Wpedantic` does generate a lot of false positives.

A middle ground would be `-Wpedantic -Wno-specific-error -Wno-specific-other-error` to disable the specific warning flags which are causing false positives after enabling `-Wpedantic`.

mwleeds:
> > I wonder if we should just add `-Wno-pedantic`?
> 
> Generally I would go with a list of known-useful warning flags, rather than `-Wpedantic`, since `-Wpedantic` does generate a lot of false positives.
> 
> A middle ground would be `-Wpedantic -Wno-specific-error -Wno-specific-other-error` to disable the specific warning flags which are causing false positives after enabling `-Wpedantic`.

I don't know what the more specific error flag would be. The compiler only mentions `-Wpedantic`.

> 
> > Looks like we also need https://gitlab.gnome.org/pwithnall/libgsystemservice/merge_requests/17
> 
> Merged, thanks.

Updated the subproject.

pwithnall:
> I don't know what the more specific error flag would be. The compiler only mentions `-Wpedantic`.

Oh, I guess you should drop `-Wpedantic` then, and list some more specific warning flags instead. There’s a list in GLib (in the top-level `meson.build` IIRC) which is fairly general purpose.

PR #52, opened 2019-09-03
Make eos-paygd use the watchdog timer

mwleeds:
https://phabricator.endlessm.com/T27051

### DIFF
--- a/eos-paygd/main.c
+++ b/eos-paygd/main.c
@@ -32,6 +32,8 @@
 #include <fcntl.h>
 #include <errno.h>
 
+#define FATAL_SIGNAL_EXIT_CODE 254
+
 /* Force a poweroff in situations where we are not able to enforce PAYG */
 static gboolean
 sync_and_poweroff (gpointer user_data)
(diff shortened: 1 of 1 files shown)


mwleeds:
Per my comment on the ticket, don't merge this yet.

ManMower:
LGTM

mwleeds:
I updated a comment to explain why we ping the watchdog even if PAYG is not active.

eos-paygd/main.c
```diff
@@ -154,7 +207,10 @@ main (int   argc,
        * initramfs difficult. */
       timeout_id = g_timeout_add_seconds (10 * 60, sync_and_poweroff, NULL);
       while (access ("/etc/initrd-release", F_OK) >= 0)
-        g_usleep (G_USEC_PER_SEC / 5);
+        {
+          g_main_context_iteration (NULL, TRUE); /* keep pinging watchdog */
+          g_usleep (G_USEC_PER_SEC / 5);
```
treitter: nitpick: I think it'd be clearer if this were expressed like `200 * USEC_PER_MSEC`

mwleeds: Honestly to me that seems more esoteric

PR #53, opened 2019-09-05
Relax daemon sandboxing

mwleeds:
https://phabricator.endlessm.com/T27037


eos-paygd/main.c
```diff
@@ -192,6 +192,12 @@ main (int   argc,
       if (system_ret == -1 || !WIFEXITED (system_ret) || WEXITSTATUS (system_ret) != 0)
         g_warning ("chown of /var/lib/eos-payg failed");
     }
+  else
+    {
+      system_ret = system ("/bin/mkdir --mode=700 /var/lib/eos-payg");
```
pwithnall: Sorry if I’m missing something, but why not just call the `mkdir()` function?

mwleeds: Honestly I just thought this would be easier since I could reuse the code above to handle the error conditions. But the man page for `mkdir()` says "The argument mode specifies the permissions to use. It is modified by the process's umask in the usual way: the permissions of the created directory are (mode & ~umask & 0777)".

Does that apply for `system()` too? Because we want it to have mode `0700` not `0700` modified by a umask.

mwleeds: Experimentally, it seems `/bin/mkdir` is unaffected by the process's umask. To me that's a good enough reason to use it even though the umask will almost certainly be `022` which would not affect `700`. I'll add a comment.

PR #57, opened 2019-09-18
eos-paygd: Exit gracefully on non-PAYG systems

mwleeds:
In case no PAYG providers are found to be enabled, EpgService takes one
of two paths: (1) in case the EOSPAYG_active EFI variable is set, the
EPG_SERVICE_ERROR_NO_PROVIDER error is returned which causes eos-paygd
to force a poweroff, since that EFI variable being set means PAYG should
be enabled, or (2) in case the EOSPAYG_active EFI variable is not set,
gss_service_exit() is called with no error set to exit from
gss_service_run(). However in this second case an error is set by
gss_service_exit() which calls g_cancellable_cancel(), so even though
the service didn't set an error there is one set for the caller of
gss_service_run(), in this case main().

So this commit checks for the G_IO_ERROR_CANCELLED error and exits with
a zero return code in that case. This means on non-PAYG systems (or PAYG
systems that have been paid off), eos-paygd.service will be considered
to have succeeded by systemd.

https://phabricator.endlessm.com/T27836


pwithnall:
Hmm, I wonder if we should handle `G_IO_ERROR_CANCELLED` inside `gss_service_run()`. If I’ve read your description properly, one of the situations is that `gss_service_run()` is called, the PAYG service decides it’s got nothing to do so calls `gss_service_exit()`, and that results in `gss_service_run()` returning `G_IO_ERROR_CANCELLED`. I suspect that `gss_service_run()` should return with no error set in that case.

What do you think?

mwleeds:
> Hmm, I wonder if we should handle `G_IO_ERROR_CANCELLED` inside `gss_service_run()`. If I’ve read your description properly, one of the situations is that `gss_service_run()` is called, the PAYG service decides it’s got nothing to do so calls `gss_service_exit()`, and that results in `gss_service_run()` returning `G_IO_ERROR_CANCELLED`. I suspect that `gss_service_run()` should return with no error set in that case.
> 
> What do you think?

Yes, I agree it would make more sense for `gss_service_run()` not to set any error when one wasn't set in the call to `gss_service_exit()`. However I'm not sure what that patch would look like. There's no obvious place in `gss_service_run()` where `@error` is being set in case the cancellable was cancelled.

mwleeds:
I'll see if I can figure out how to patch libgsystemservice; that seems like the best way forward in terms of code quality.

mwleeds:
https://gitlab.gnome.org/pwithnall/libgsystemservice/merge_requests/18

pwithnall:
[libgsystemservice change](https://gitlab.gnome.org/pwithnall/libgsystemservice/merge_requests/18/diffs) merged upstream, so this PR needs updating to bump the submodule version to incorporate it…and I think that should be it?

mwleeds:
Done, thanks

PR #58, opened 2019-09-20
eos-paygd: chroot before the root pivot

ManMower:
We need to chroot into /sysroot before the root pivot occurs so
the dbus socket (among many other things) is visible to us. But
we have to do that *after* telling system it's ok to proceed with
the pivot, and start deleting the root of our current directory
hierarchy.

Do this with a special sd_notify variant that lets us chroot
away from the space the sd notification socket lives in before
sending the notification.

Also, to make this chroot possible, relax a bunch of sandboxing
settings.

https://phabricator.endlessm.com/T27035


ManMower:
Sigh, typo in the ticket number.  taking a mulligan.

PR #59, opened 2019-09-20
eos-paygd: chroot before the root pivot

ManMower:
We need to chroot into /sysroot before the root pivot occurs so
the dbus socket (among many other things) is visible to us. But
we have to do that *after* telling system it's ok to proceed with
the pivot, and start deleting the root of our current directory
hierarchy.

Do this with a special sd_notify variant that lets us chroot
away from the space the sd notification socket lives in before
sending the notification.

Also, to make this chroot possible, relax a bunch of sandboxing
settings.

https://phabricator.endlessm.com/T27037

### DIFF
--- a/eos-paygd/main.c
+++ b/eos-paygd/main.c
@@ -31,9 +31,13 @@
 #include <glob.h>
 #include <fcntl.h>
 #include <errno.h>
+#include <sys/types.h>
+#include <sys/socket.h>
+#include <sys/un.h>
 
 #define FATAL_SIGNAL_EXIT_CODE 254
 #define WATCHDOG_FAILURE_EXIT_CODE 253
+#define SD_NOTIFY_FAILURE_EXIT_CODE 252
 
 static int watchdog_fd = -1;
 
@@ -131,6 +135,44 @@ allow_writing_to_boot_partition (gboolean allow_write)
   globfree (&globbuf);
 }
 
+static int
+payg_relative_sd_notify (const gchar *path,
+                         const char *state)
+{
+  struct sockaddr_un sockaddr = {
+    .sun_family = AF_UNIX,
+  };
+  struct iovec iovec = {
+    .iov_base = (char *)state,
+    .iov_len = strlen(state),
+  };
+  struct msghdr msghdr = {
+    .msg_iov = &iovec,
+    .msg_iovlen = 1,
+    .msg_name = &sockaddr,
+  };
+  int fd;
+  int l, r = 1, salen;
+
+  unsetenv("NOTIFY_SOCKET");
+
+  l = strlen(path);
+  memcpy(&sockaddr.sun_path, path, l + 1);
+  salen = offsetof(struct sockaddr_un, sun_path) + l + 1;
+
+  fd = socket(AF_UNIX, SOCK_DGRAM|SOCK_CLOEXEC, 0);
+  if (fd < 0)
+    return -errno;
+
+  msghdr.msg_namelen = salen;
+  if (sendmsg(fd, &msghdr, MSG_NOSIGNAL) < 0)
+    r = -errno;
+
+  close(fd);
+
+  return r;
+}
+
 int
 main (int   argc,
       char *argv[])
@@ -141,6 +183,9 @@ main (int   argc,
   int ret, sd_notify_ret, system_ret;
   gboolean backward_compat_mode = FALSE;
   guint timeout_id, watchdog_id;
+  const gchar *sd_socket_env = NULL;
+  g_autofree char *sd_socket_dir = NULL;
+  g_autofree char *sd_socket_name = NULL;
 
   /* If eos-paygd is running from the initramfs, change the process name so
    * that it survives the pivot to the final root filesystem. This is an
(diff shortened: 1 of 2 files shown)


eos-paygd/main.c
```diff
@@ -194,24 +239,35 @@ main (int   argc,
...
+       * doesn't refuse to use relative paths to notify systemd we're ready.
+       *
+       * After that, we can chdir() into our new / and begin trying to connect
+       * to the dbus socket. */
+
+      sd_socket_env = g_getenv("NOTIFY_SOCKET");
+      sd_socket_dir = g_path_get_dirname(sd_socket_env);
+      sd_socket_name = g_path_get_basename(sd_socket_env);
+      chdir(sd_socket_dir);
+      chroot("/sysroot");
```
mwleeds: Is `/sysroot` guaranteed to exist yet? My understanding is that gets created at some point while the initrd is running but is it guaranteed to be before `eos-payg-2.service` starts? We only have `After=sysinit.target` but I think we need `After=sysroot.mount` if I'm reading this correctly: https://www.freedesktop.org/software/systemd/man/bootup.html#

eos-paygd/main.c
```diff
@@ -194,24 +239,35 @@ main (int   argc,
...
+       * pivot.
+       *
+       * What we do is chdir() into the directory $NOTIFY_SOCKET resides in,
+       * chroot() into /sysroot, then use our own variant of sd_notify() that
+       * doesn't refuse to use relative paths to notify systemd we're ready.
+       *
+       * After that, we can chdir() into our new / and begin trying to connect
+       * to the dbus socket. */
+
+      sd_socket_env = g_getenv("NOTIFY_SOCKET");
```
mwleeds: Should we have some sanity check here so if `$NOTIFY_SOCKET` is unset for some reason, we print a warning and don't try to use the value?

eos-paygd/main.c
```diff
@@ -194,24 +239,35 @@ main (int   argc,
...
+      sd_socket_dir = g_path_get_dirname(sd_socket_env);
+      sd_socket_name = g_path_get_basename(sd_socket_env);
+      chdir(sd_socket_dir);
```
mwleeds: This `chdir`, the `chroot`, and the second `chdir` could also do with having their return values checked and a warning printed on error.

eos-paygd/main.c
ManMower: The /sysroot mount point is actually a dir in the dracut image, so it's technically always there... and since we open the dbus socket with an absolute path, that will work.  It's probably a good idea to chdir("/"); yet one more time after we successfully connect to dbus, so relative paths will work - we'll probably be "under" the root mount for relative paths until I add  that..

ManMower: Newly pushed changes add the chdir("/") after dbus connect

eos-paygd/main.c
ManMower: I've added checks to all of this stuff.  I explicitly exit when NOTIFY_SOCKET is unset or bogus, as that just leads to a drawn out and inescapable wait for shutdown anyway.

The rest print warnings, as there's a chance we'll actually make it to user space and be able to do something until the watchdog timer gets us.

ManMower:
most recent forced push was just to make the remaining sd_notify failure that didn't error out be consistent with the others.

mwleeds:
Typo in the commit message: "system" -> "systemd"

eos-paygd/main.c
```diff
@@ -194,24 +240,52 @@ main (int   argc,
...
+       * After that, we can chdir() into our new / and begin trying to connect
+       * to the dbus socket. */
+
+      sd_socket_env = g_getenv("NOTIFY_SOCKET");
+      if (!sd_socket_env)
+        {
+          /* If we can't notify systemd that we're ready to move on, it will
+           * timeout in 90 seconds and shutdown, might as well just exit. */
+          g_printerr ("%s: NOTIFY_SOCKET not set.\n", argv[0]);
+          return SD_NOTIFY_FAILURE_EXIT_CODE;
```
mwleeds: By returning early we're skipping the `allow_writing_to_boot_partition (FALSE);` below. I guess that's not an issue; the boot partition is writable while untrusted userspace code is running anyway. Just wanted to make sure we're aware.

eos-paygd/main.c
```diff
@@ -233,7 +307,8 @@ main (int   argc,
         }
       g_source_remove (timeout_id);
     }
-
+  if (chdir ("/"))
```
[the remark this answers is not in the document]
mwleeds: Also worth adding a comment to explain why this is needed?

eos-paygd/main.c
```diff
@@ -194,24 +240,52 @@ main (int   argc,
...
+        {
+          g_printerr ("%s: NOTIFY_SOCKET not in valid format.\n", argv[0]);
+          return SD_NOTIFY_FAILURE_EXIT_CODE;
+        }
+      if (chdir (sd_socket_dir))
+        g_warning ("Unable to change working dir to systemd socket dir (%s): %m", sd_socket_dir);
+      if (chroot ("/sysroot"))
+        g_warning ("Unable to switch root to run-time root directory (/sysroot): %m");
+
+      sd_notify_ret = payg_relative_sd_notify (sd_socket_name, "READY=1");
```
mwleeds: Is the socket we're using here the same one as in the initrd filesystem? And if not, how do we know systemd will have already created it in the new sysroot?

It's possible that's a stupid question because I don't have a good conceptual grasp on how sockets work.

eos-paygd/main.c
ManMower: Forgot to resolve this one, there's now a sanity check.

eos-paygd/main.c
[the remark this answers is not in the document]
ManMower: More importantly, this code is in the wrong place by one }  :)

Oops.

eos-paygd/main.c
ManMower: Same situation exists for the watchdog timer, but I think at this point an exit with failure code is going to result in an immediate shutdown while we're still in early initrd?

eos-paygd/main.c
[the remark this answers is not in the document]
ManMower: Added a comment, and moved the offending code block under the correct conditional in recent force push.

eos-paygd/main.c
ManMower: yes, the socket we're using is the one in initramfs (exactly what systemd told us to use)

essentially something like this:
NOTIFY_SOCKET=/var/run/systemd/notify
we cd /var/run/systemd
we use "notify" as the socket name.

eos-paygd/main.c
mwleeds: You're right we do have `FailureAction=poweroff`

eos-paygd/main.c
mwleeds: Okay. It feels a bit fragile since maybe a process having a cwd outside its root directory is a rare situation that may be broken by a kernel update. But I don't have a better idea.

PR #60, opened 2019-09-23
fix build on arm

ManMower:
We have a 64 bit integer in one struct definition, and we're casting
another struct that contains nothing that large to it. The struct
that only contains smaller types only has a 32-bit alignment
constraint, but is being cast to something with a 64-bit alignment
constraint, so gcc complains.

Break the 64-bit item into high/low components and recombine/break
apart as required.

### DIFF
--- a/libeos-payg/fake-clock.c
+++ b/libeos-payg/fake-clock.c
@@ -110,15 +110,20 @@ typedef struct {
   GSource parent;
 
   EpgFakeClock *clock; /* owned */
-  gint64 ready_time_secs;
+  guint32 ready_time_secs_high;
+  guint32 ready_time_secs_low;
 } EpgFakeSource;
 
 static gboolean
 epg_fake_source_check (GSource *source)
 {
+  gint64 ready_time_secs;
   EpgFakeSource *self = (EpgFakeSource *) source;
 
-  return epg_clock_get_time (EPG_CLOCK (self->clock)) >= self->ready_time_secs;
+  ready_time_secs = self->ready_time_secs_high;
+  ready_time_secs = (ready_time_secs << 32) +
+                    self->ready_time_secs_low;
+  return epg_clock_get_time (EPG_CLOCK (self->clock)) >= ready_time_secs;
 }
 
 static gboolean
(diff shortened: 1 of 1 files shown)


libeos-payg/fake-clock.c
```diff
@@ -110,15 +110,20 @@ typedef struct {
...
 
 static gboolean
 epg_fake_source_check (GSource *source)
 {
+  gint64 ready_time_secs;
   EpgFakeSource *self = (EpgFakeSource *) source;
 
-  return epg_clock_get_time (EPG_CLOCK (self->clock)) >= self->ready_time_secs;
+  ready_time_secs = self->ready_time_secs_high;
+  ready_time_secs = (ready_time_secs << 32) +
```
mwleeds: This is not quite correct if `self->ready_time_secs_high == G_MAXUINT32` right? Since won't `ready_time_secs` use one bit to store its sign?

And could we do it in one line and use bitwise OR: `ready_time_secs = (gint64) (self->ready_time_secs_high << 32) | self->ready_time_secs_low`

ManMower: I don't see the problem - ready_time_secs is indeed 63-bits of number and 1-bit of sign, but bit shifting left doesn't care about signs at all, and while bit shifting right sign extends, I don't see how that can hurt us.  Am I missing something?

In your other suggestion you would need to cast inside the braces to avoid incorrect computation (and a compiler warning).

| and + should give exactly the same answer here because _low is unsigned and being combined with nothing but 0s.

Is there really a concern about the number of lines of code this takes? I find it more obvious what's happening without having to deal with explicit typecasting.

mwleeds: I was worried that If `ready_time_secs` were somehow negative, then `ready_time_secs_high = ready_time_secs >> 32;` would not set the leftmost bit of `ready_time_secs_high` if the compiler noticed that it's assigning to an unsigned integer. But experimentally from running [this](https://gist.github.com/mwleeds/0437d23a6c28d47c1d6e3e8a77bf39b2), it appears this does work.

mwleeds: Never mind about the one-line solution. Let's keep what you have, especially after I've tested it independently.

PR #61, opened 2019-10-03
WIP: fix Meson warnings

wjt:
I spotted these while doing a round of translations on https://phabricator.endlessm.com/T27764.

TODO:

- [ ] Verify that the tests pass. I've only run `meson _build`.
- [ ] Fix the remaining warning below:

```
manual-provision-payg/meson.build:1: WARNING: Custom target input 'testsuite' can't be converted to File object(s).
This will become a hard error in the future.
```

but I don't plan to work further on this branch.

### DIFF
--- a/eos-payg-csv/tests/meson.build
+++ b/eos-payg-csv/tests/meson.build
@@ -1,5 +1,5 @@
-py3_mod = import('python3')
-py3 = py3_mod.find_python()
+python_mod = import('python')
+py3 = python_mod.find_installation('python3')
 
 envs = test_env + [
   'G_TEST_SRCDIR=' + meson.current_source_dir(),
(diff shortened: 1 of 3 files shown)


mwleeds:
> I spotted these while doing a round of translations on https://phabricator.endlessm.com/T27764.
> 
> TODO:
> 
>     * [ ]  Verify that the tests pass. I've only run `meson _build`.
They do according to Jenkins
> 
>     * [ ]  Fix the remaining warning below:
> 
> 
> ```
> manual-provision-payg/meson.build:1: WARNING: Custom target input 'testsuite' can't be converted to File object(s).
> This will become a hard error in the future.
> ```

This is because `testsuite` is a directory rather than a file, but per [this issue](https://github.com/mesonbuild/meson/issues/4717) upstream there's not a great solution yet. I don't think we need to block merging these fixes on fixing that warning.
> 
> but I don't plan to work further on this branch.

mwleeds:
LGTM

PR #63, opened 2019-10-07
eos-paygd: Don't enforce PAYG if Secure Boot is off

mwleeds:
If the current boot occurred with Secure Boot disabled, it doesn't make
sense to enforce PAYG because (a) it's easily circumventable by
replacing the initramfs and (b) Secure Boot has likely been disabled
because the machine is in one of two states:
1. It has been fully paid off and unlocked (though note that it's also
possible to have Secure Boot enabled on an unlocked machine, in which
case EOSPAYG_active will not be set).
2. It is being debugged by a developer who has used PKpriv to disable
Secure Boot.

This commit also adds a command line flag --skip-sb-check, so a
developer can have eos-paygd in enforcing mode even with Secure Boot
off (though adding the flag would require editing the initramfs or using
a specially created debug image).

https://phabricator.endlessm.com/T27735

### DIFF
--- a/eos-paygd/main.c
+++ b/eos-paygd/main.c
@@ -41,6 +41,8 @@
 #define SD_NOTIFY_FAILURE_EXIT_CODE 252
 #define GOPTION_FAILURE_EXIT_CODE 251
 
+#define EFI_GLOBAL_VARIABLE_GUID EFI_GUID(0x8be4df61, 0x93ca, 0x11d2, 0xaa0d, 0x00, 0xe0, 0x98, 0x03, 0x2b, 0x8c)
+
 static int watchdog_fd = -1;
 
 /* Ping the watchdog periodically as long as eos-paygd is running. */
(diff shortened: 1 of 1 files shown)


eos-paygd/main.c
```diff
@@ -231,11 +233,36 @@ test_and_update_securitylevel (char *procname)
...
+static gboolean
+secure_boot_enabled (void)
+{
+  uint8_t *secboot = NULL;
+  size_t data_size = 0;
+  uint32_t attributes;
+  int ret;
+
+  ret = efi_get_variable (EFI_GLOBAL_VARIABLE_GUID, "SecureBoot", &secboot, &data_size, &attributes);
+  if (ret < 0 || !secboot || data_size != 1)
```
ManMower: !secboot just means secure boot is off here, doesn't it? It should be left to the other conditional

eos-paygd/main.c
```diff
@@ -231,11 +233,36 @@ test_and_update_securitylevel (char *procname)
...
+  return TRUE;
+}
+
 static gboolean print_level = FALSE;
+static gboolean skip_sb_check = FALSE;
 
 static GOptionEntry opts[] =
 {
   { "seclevel", 's', 0, G_OPTION_ARG_NONE, &print_level, "Print security level and exit", NULL },
+  { "skip-sb-check", 0, 0, G_OPTION_ARG_NONE, &print_level, "Enforce PAYG even if Secure Boot is off", NULL },
```
ManMower: s/print_level/skip_sb_check/

eos-paygd/main.c
mwleeds: No, looking at the implementation in libefivar, I think it's just never the case that `ret == 0 && secboot == NULL`. And experimentally my machine has SB off and on it `secboot != NULL && *secboot == 0`. Side note: I changed the conditional below from `secboot == 0` to `*secboot == 0`.

eos-paygd/main.c
mwleeds: Oops

eos-paygd/main.c
ManMower: Oops, you're right, the other check was the bug, and you've fixed it.

PR #64, opened 2019-10-08
eos-paygd: Only ping watchdog every minute

mwleeds:
Currently we ping the watchdog timer every 10 seconds, but this is a bit
excessive given that it's a 20 minute timer. So increase the time
between pings to 60 seconds, and update the documentation to match the
changes made here[1]. I'm not sure exactly what the performance cost is
of this ping, which should just be in memory. We do know the timer stops
when the computer is suspended, so it's okay that pings won't happen
in that state.

[1] https://github.com/endlessm/linux/pull/502

### DIFF
--- a/eos-paygd/main.c
+++ b/eos-paygd/main.c
@@ -358,10 +358,10 @@ main (int   argc,
   if (!backward_compat_mode)
     {
       /* Open and start pinging the custom watchdog timer ("endlessdog") which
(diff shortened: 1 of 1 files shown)


mwleeds:
I noticed the corresponding kernel change for this isn't on the eos3.7 branch, so I created this ticket for backporting this and that: https://phabricator.endlessm.com/T28062

PR #65, opened 2019-10-08
Shut down after 10 minutes when there's no credit remaining

mwleeds:
https://phabricator.endlessm.com/T27133


mwleeds:
Sadly the existing tests don't seem to use `EpgManagerService` so I don't think there's a way to add a unit test for this that's a reasonable amount of work. But I added a comment about adding this to our test suite here: https://phabricator.endlessm.com/T27070

mwleeds:
Merging this since it's been reviewed and https://github.com/endlessm/gnome-shell/pull/539 has landed.

PR #68, opened 2019-10-28
Implement support for EOSPAYG_debug EFI variable

mwleeds:
This EFI variable should prove useful when debugging PAYG systems which
in production are so locked down as to be difficult to debug. You can,
for example, set a bit to tell eos-paygd not to use the LSM, which means
the process will show up in /proc and can be attached to with gdb. This
variable cannot easily be set on a production system because the LSM
ensures that only eos-paygd can access EFI variables which have the
prefix "EOSPAYG". The expected flow is that the developer will build a
bootable EFI blob which unlocks the computer in question if the account
number on the computer matches the one compiled into the blob. See
https://phabricator.endlessm.com/T28290. Then, even if the primary OS
isn't booting, you could disable Secure Boot, boot from a USB drive, and
set EOSPAYG_debug using the instructions here:
https://phabricator.endlessm.com/w/kernel/efivars/

https://phabricator.endlessm.com/T27734

### DIFF
--- a/libeos-payg/util.c
+++ b/libeos-payg/util.c
@@ -24,6 +24,9 @@
 #include <gio/gio.h>
 #include <linux/reboot.h>
 #include <sys/reboot.h>
+#include <efivar.h>
+
+#define EFI_GLOBAL_VARIABLE_GUID EFI_GUID(0x8be4df61, 0x93ca, 0x11d2, 0xaa0d, 0x00, 0xe0, 0x98, 0x03, 0x2b, 0x8c)
 
 /* Force a poweroff in situations where we are not able to enforce PAYG. This
  * is intended to be used as a GSourceFunc, e.g. with g_timeout_add_seconds()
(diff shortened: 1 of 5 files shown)


mwleeds:
We should hold off on merging this until I've tested it, which should be tomorrow.

libeos-payg/util.c
```diff
@@ -0,0 +1,247 @@
...
+
+  /* The C standard doesn't guarantee the underlying type of an enum, so let's
+   * convert manually to be safe.
+   */
+  if (debug_efivar[0] & EPG_DEBUG_SET_G_MESSAGES_DEBUG_ALL)
+    debug_flags |= EPG_DEBUG_SET_G_MESSAGES_DEBUG_ALL;
+  if (debug_efivar[0] & EPG_DEBUG_SECURE_BOOT_OFF)
+    debug_flags |= EPG_DEBUG_SECURE_BOOT_OFF;
+  if (debug_efivar[0] & EPG_DEBUG_SECURE_BOOT_ON)
+    debug_flags |= EPG_DEBUG_SECURE_BOOT_ON;
```
ManMower: This is exactly what I wanted to see, where secure boot can be forced on, or force off, or just left alone.

libeos-payg/util.c
```diff
@@ -0,0 +1,247 @@
...
+  uint8_t *secboot = NULL;
+  size_t data_size = 0;
+  uint32_t attributes;
+  int ret;
+
+  if (debug_flags & EPG_DEBUG_SECURE_BOOT_OFF &&
+      debug_flags & EPG_DEBUG_SECURE_BOOT_ON)
+    g_printerr ("%s: Both EPG_DEBUG_SECURE_BOOT_OFF and EPG_DEBUG_SECURE_BOOT_ON are set\n", G_STRFUNC);
+
+  if (debug_flags & EPG_DEBUG_SECURE_BOOT_OFF)
```
ManMower: Optional pedantry - would it be useful to know if our override didn't actually override state?
ie) secure boot was already off but we forced it off

mwleeds: That would require reading EFI variables just for the sake of a debug message. I think that's not worth it, in case there's a price to pay in terms of performance or wear on the flash memory.

mwleeds:
I fixed the merge conflicts and tested that `EPG_DEBUG_SECURE_BOOT_ON` behaves as advertised so I think this is mergeable. The other bits can be tested during QA.

mwleeds:
Force-pushed this so we initialize `watchdog_id` to `0` (and just for peace of mind also initialize `timeout_id`).

libeos-payg/util.h
```diff
@@ -16,13 +16,21 @@
  * LGPL. If you do not delete the provisions above, a recipient may use your
  * version of this file under the terms of either the MPL or the LGPL.
  */
-
```
ManMower: I suppose this stands out as an unrelated (unintentional?) formatting change.

I'm not really bothered by it.

ManMower:
I'm embarrassed - I remember making a mental note to check watchdog_id's provenance when I saw its new usage. Apparently I didn't follow up on that at all.

Approval still stands.  good catch.

libeos-payg/util.h
mwleeds: Fixed this

PR #71, opened 2019-11-04
Add and use EpgProvider::unlocked signal

mwleeds:
Currently if one enters a final unlock code which pays off a computer's
loan, the EpgProvider object sets `enabled` to FALSE and sets the expiry
time far into the future, so that PAYG is no longer enforced. However,
since the EOSPAYG_active EFI variable is left set on Phase 4+ systems,
upon the next boot, eos-paygd will initiate a shutdown since it doesn't
expect state data to be in a "PAYG disabled" state while EOSPAYG_active
is set. Thus the user can't really use their paid off computer for more
than a few minutes at a time.

So this commit adds an `unlocked` signal to EpgProvider, which will be
emitted by the Angaza backend when the final unlock code has been
entered. The signal emission is caught by eos-paygd which deletes
EOSPAYG_active, so the computer will be free to use on the next boot
(and for the current boot as well).

An alternative implementation would be to use the existing callback
that's triggered whenever the `enabled` property changes, and do the
EOSPAYG_active deletion there. But that seems fragile to me, since a
provider is also considered disabled if it has not been provisioned
(e.g. the provider is for Paygee and the computer was provisioned for
Angaza) whereas a computer is only ever unlocked once. So, add a new
signal to match the expected semantics and make the code more
maintainable.

It doesn't make sense to implement this for EpgManager which is
deprecated, so there's also no way to test it in this repo. But we can
test it in the tests for the Angaza backend.

https://phabricator.endlessm.com/T28461

### DIFF
--- a/libeos-payg/service.c
+++ b/libeos-payg/service.c
@@ -34,6 +34,7 @@
 #include <libeos-payg-codes/codes.h>
 #include <libgsystemservice/config-file.h>
 #include <locale.h>
+#include <efivar.h>
 
 
 /* Paths to the various places the config file could be loaded from. */
@@ -96,6 +97,11 @@ struct _EpgService
    */
   gulong notify_enabled_id;
 
+  /* If @provider is non-%NULL, a non-zero ID for a handler of
+   * provider::unlocked.
+   */
+  gulong unlocked_id;
+
   /* If %TRUE, there is an outstanding call to gss_service_hold() without a
    * matching call to gss_service_release().
    */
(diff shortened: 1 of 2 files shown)


libeos-payg/service.c
```diff
@@ -526,6 +546,30 @@ provider_notify_enabled_cb (GObject    *object,
...
+   * EOSPAYG_active (via EPG_DEBUG_EOSPAYG_ACTIVE_OFF).
+   */
+  if (self->eospayg_active_efivar)
+    {
+      if (efi_del_variable (EOSPAYG_GUID, "EOSPAYG_active") < 0)
+        g_warning ("Failed to delete EOSPAYG_active upon unlock");
+      else
+        g_message ("Deleted EOSPAYG_active upon unlock");
+
+      /* FIXME: Perhaps we should also change the Secure Boot settings so the
```
treitter: Let's cut this for now just to reduce our testing burden. But I agree we should follow up on it soon so I filed https://phabricator.endlessm.com/T28511 . (I thought we already had a task for this but I couldn't find it)

mwleeds: I don't see how a code comment increases testing burden?

mwleeds: I added that phabricator link to the comment. I guess you meant our burden would be increased if we implemented that change.

PR #73, opened 2019-11-14
eos-payg: Add the account-id property

mazen-asef:
Add an account-id property to be shown on the user's unlock
screen every time he adds a Pay-As-You-Go credit. As of now, we
don't assign a proper code to the user, so this is mostly a
foundation for the eos-payg-nonfree module.

https://phabricator.endlessm.com/T28491

### DIFF
--- a/libeos-payg/errors.h
+++ b/libeos-payg/errors.h
@@ -32,6 +32,8 @@ G_BEGIN_DECLS
  * @EPG_MANAGER_ERROR_TOO_MANY_ATTEMPTS: Too many attempts to verify a code
  *    in recent history.
  * @EPG_MANAGER_ERROR_DISABLED: Pay as you go is disabled.
+ * @EPG_MANAGER_ERROR_DISPLAY_ACCOUNT_ID: The given code was valid, but instead of
+ *    adding time, tell the listener to show the account ID assigned to the machine.
  *
  * Errors which can be returned by #EpgManager.
  *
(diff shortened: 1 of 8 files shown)


libeos-payg/errors.h
```diff
@@ -32,6 +32,8 @@ G_BEGIN_DECLS
  * @EPG_MANAGER_ERROR_TOO_MANY_ATTEMPTS: Too many attempts to verify a code
  *    in recent history.
  * @EPG_MANAGER_ERROR_DISABLED: Pay as you go is disabled.
+ * @EPG_MANAGER_ERROR_DISPLAY_ACCOUNT_ID: The given code was valid, but instead of
+ * adding time, tell the listener to show the account ID assigned to the machine.
```
mwleeds: indentation is off here

libeos-payg/manager.c
```diff
@@ -438,6 +448,11 @@ epg_manager_set_property (GObject      *object,
       g_assert (self->clock == NULL);
       self->clock = g_value_dup_object (value);
       break;
+    case PROP_ACCOUNT_ID:
+      /* Construct only. */
+      g_assert (self->account_id == NULL);
+      self->account_id = g_value_get_string (value);
```
mwleeds: I don't think you want to allow setting it.

libeos-payg/tests/plugins/test-provider.c
```diff
@@ -212,6 +220,11 @@ epg_test_provider_set_property (GObject      *object,
       g_assert (priv->clock == NULL);
       priv->clock = g_value_dup_object (value);
       break;
+    case PROP_ACCOUNT_ID:
+      /* Construct only. */
+      g_assert (priv->account_id == NULL);
+      priv->account_id = g_value_get_string (value);
```
mwleeds: As above, this shouldn't be here.

libeos-payg/errors.h
mwleeds: Also, I think you should update the docs for `epg_provider_add_code()` to mention that this error can be returned and should be handled.

PR #75, opened 2020-01-31
Add util function for checking if this is a legacy or modern PAYG system

ManMower:
We need a way for plugins to check if they should use modern or legacy
code paths. This will allow us to, for example, disable datastore
encryption on phase 2 deployments.

https://phabricator.endlessm.com/T27052

### DIFF
--- a/libeos-payg/util.c
+++ b/libeos-payg/util.c
@@ -28,6 +28,8 @@
 
 #define EFI_GLOBAL_VARIABLE_GUID EFI_GUID(0x8be4df61, 0x93ca, 0x11d2, 0xaa0d, 0x00, 0xe0, 0x98, 0x03, 0x2b, 0x8c)
 
+static gboolean payg_legacy_mode = FALSE;
+
 /* Force a poweroff in situations where we are not able to enforce PAYG. This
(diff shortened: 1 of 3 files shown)


libeos-payg/util.c
```diff
@@ -263,3 +265,26 @@ payg_should_check_securitylevel (void)
 
   return TRUE;
 }
+
+/**
+ * payg_get_legacy_mode:
+ *
+ * If eospaygd is running in Phase 2 more, returns %TRUE, otherwise returns
```
mwleeds: Typo: mode

libeos-payg/util.c
```diff
@@ -263,3 +265,26 @@ payg_should_check_securitylevel (void)
...
+{
+  return payg_legacy_mode;
+}
+
+/**
+ * payg_internal_set_legacy_mode:
+ *
+ * To be called by eospaygd to initialize the value returned by
+ * payg_get_legacy_mode. Should be called once, only on Phase 2
+ * systems.
```
mwleeds: Might be worth adding a parenthetical at the end "(where eos-paygd is run from the primary filesystem not the initramfs)" in case someone reading this is confused about what Phase 2 means.

PR #79, opened 2020-08-31
efi: Add mini efi library to replace efivar

ManMower:
Due to the unique nature of our efi variable access patterns, libefivar
isn't currently suited to our needs.

For now, replace efivar with our own tiny library to deal with efivarfs.
This also allows us to simplify API to fit our requirements, and make it
more resilient to things we consider bugs in eos-paygd.

There are future plans (T30696, T30698) to reintroduce the libefivar
dependency when we have time to address our needs in upstream friendly
ways.

https://phabricator.endlessm.com/T25224

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -14,7 +14,6 @@ Build-Depends:
  meson,
  python3-dbusmock,
  dracut,
- libefivar-dev,
  gtk-doc-tools,
 
 Package: eos-paygd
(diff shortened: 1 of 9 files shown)


ManMower:
This code may see some small changes as I finish up T25224, so please don't feel an urgent need to review right away. But even broad comments could be helpful.

Thanks!

ManMower:
forced push just bumps the number of variables we can store in the fake test-mode storage implementation from 50 to 200.

New version of the test script makes more than 64 at a time, which was over the old limit.

Still too lazy to make the limit dynamic, 200 variable should be enough for anyone.

PR #81, opened 2020-10-01
eos-paygd: Increase security level

ManMower:
Enough has changed this release that we've decided to bump the
security level.

https://phabricator.endlessm.com/T30900


ManMower:
quick forced push to make the commit log more clear as to why we're bumping the level.

PR #84, opened 2020-11-27
efi: Add a function to check for SecureBootOption

ManMower:
This variable is non-standard and weird, but in the cases it
exists it might be useful to query.

https://phabricator.endlessm.com/T31137

### DIFF
--- a/libeos-payg/efi.c
+++ b/libeos-payg/efi.c
@@ -18,6 +18,8 @@
 #define EOSPAYG_GUID         "d89c3871-ae0c-4fc5-a409-dc717aee61e7"
 #define GLOBAL_VARIABLE_GUID "8be4df61-93ca-11d2-aa0d-00e098032b8c"
 
+/* Can't figure out who owns this GUID. ECS and Lenovo have both used it */
+#define SBO_VARIABLE_GUID    "955b9041-133a-4bcf-90d1-97e1693c0e30"
 #define NVM_PREFIX           "EOSPAYG_"
 
 /* Totally bogus low performance mock storage for dry runs.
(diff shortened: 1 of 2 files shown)


libeos-payg/efi.c
```diff
@@ -474,6 +484,65 @@ eospayg_efi_secureboot_active (void)
...
 
+/* eospayg_efi_securebootoption_disabled:
+ *
+ * Check if the SecureBootOption EFI variable exists and
+ * is disabled.
+ *
+ * The oddly inverted logic is due to the fact that most
+ * systems don't have this variable at all - if it doesn't
+ * exist, we can infer nothing about the state of Secure Boot,
+ * if it does it tells us if Secure Boot is on or off.
```
jprvita, line 495: We should probably try to be extra clear here and go with something like:
```suggestion
 * if it does it tells us if the Secure Boot option in the
 * BIOS is enabled or disabled.
```

libeos-payg/efi.c
```diff
@@ -474,6 +484,65 @@ eospayg_efi_secureboot_active (void)
...
+ *
+ * Thus, on and not existing should likely be treated in the
+ * same way by a caller, but existing and off is a red flag
+ * for PAYG enforcement.
+ *
+ * Returns: %TRUE if the variable exists and is disabled, %FLASE otherwise
```
jprvita, line 501:
```suggestion
 * Returns: %TRUE if the variable exists and is disabled, %FALSE otherwise
```

jprvita (changes requested):
Left a small comment suggestion and a typo fix, but otherwise LGTM.

ManMower:
forced push makes the requested changes. thanks!

jprvita (approved):

PR #85, opened 2020-12-14
libeos-payg: Try a clean shutdown first

ManMower:
Instead of a sync and a hard shutdown, we'll try telling systemd to shut
down cleanly. If that fails, or fails to happen in a reasonable amount
of time, a hard shutdown will follow.

https://phabricator.endlessm.com/T31166

### DIFF
--- a/libeos-payg/util.c
+++ b/libeos-payg/util.c
@@ -31,13 +31,24 @@ static gboolean payg_legacy_mode = FALSE;
 /* Force a poweroff in situations where we are not able to enforce PAYG. This
  * is intended to be used as a GSourceFunc, e.g. with g_timeout_add_seconds()
  */
-gboolean
+__attribute__ ((__noreturn__)) gboolean
 payg_sync_and_poweroff (gpointer user_data)
 {
   g_message ("bcsn: Forcing poweroff now!");
-  sync ();
-  reboot (LINUX_REBOOT_CMD_POWER_OFF);
-  return G_SOURCE_REMOVE;
+  if (system ("systemctl poweroff") != 0)
+    {
+      /* If we're here, something went wrong in the call, so we can force
+       * an immediate dirty shutdown, since we know a clean shutdown has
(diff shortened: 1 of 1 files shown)


libeos-payg/util.c
```diff
@@ -31,13 +31,24 @@ static gboolean payg_legacy_mode = FALSE;
...
+       * already failed.
+       */
+      sync ();
+      reboot (LINUX_REBOOT_CMD_POWER_OFF);
+    }
+  /* We've requested a shutdown, so there's nothing left for us to do.
+   * Exit now so the watchdog timer will make sure we get our shutdown
+   * eventually.
+   */
+  exit (EXIT_FAILURE);
```
jprvita, line 51: Is this really a failure? I understand we want to exit so if `systemctl poweroff` fails, the watchdog eventually kicks in, but this seems like normal exit to me.

jprvita (changes requested):
Looks good, just one question on the exit status during regular poweroff.

libeos-payg/util.c
ManMower: This is called for eos-paygd failures that are so bad that we have to shutdown the computer in response, I don't think of those are normal exit conditions.

The current callers are: failure to access EFI variables, failure of the security level monotonicity test, failure to init the rtc clock code, failure to connect to dbus, failure to read PAYG NV store and failure to enter a timecode in a reasonable amount of time.

That said, I'm pretty certain nothing's actually checking that exit code, so I'm not going to complain too much if you'd rather see it be EXIT_SUCCESS.

jprvita: I was thinking of the "credit expired" situation as a normal one, but I guess you are right, it is not "credit expired" that is triggering the shutdown in that case, it is "the computer has been running for 10 minutes after credit has expired". Let's stick with EXIT_FAILURE then.

jprvita (approved):

PR #86, opened 2021-02-01
manual-provision: Remove provisioning tarballs

ManMower:
These are obsolete now.

https://phabricator.endlessm.com/T31277


jprvita:
Rebased on the latest master.

This PR removes dead code that I don't think we are ever going to use again. However, IDK if merging this now will trigger automatically buying new translations for no reason on the next release, so maybe we should wait until we have meaningful changes we want in a release? @wjt Thoughts?

wjt:
We never automatically *buy* translations. That's always a manual step.

Source strings from the stable branch (currently `eos3.9`) are periodically pushed to Transifex, and translated strings are periodically pulled from Transifex to the stable branch, currently `eos3.9`, and to `master` if the two branches are level.

Community translators may translate stuff at any time.

This PR makes no changes to translatable strings and is targeting `master`.

PR #92, opened 2022-01-18
paygd: Request a poweroff on SIGTERM / SIGINT

jprvita:
If the daemon receives one of these termination signals we should
trigger a poweroff, so the computer does not continue to run without
paygd.

This is required for us to allow the daemon to be terminated by systemd
during shutdown but not manually by the user.

https://phabricator.endlessm.com/T33088
https://phabricator.endlessm.com/T32815

### DIFF
--- a/eos-paygd/main.c
+++ b/eos-paygd/main.c
@@ -40,6 +40,7 @@
 #define SD_NOTIFY_FAILURE_EXIT_CODE 252
 #define GOPTION_FAILURE_EXIT_CODE 251
 #define LSM_FAILURE_EXIT_CODE 250
+#define NO_PROVIDER_EXIT_CODE 249
 
 static int watchdog_fd = -1;
 
@@ -467,6 +468,7 @@ main (int   argc,
             break;
         }
       g_source_remove (timeout_id);
+      timeout_id = 0;
 
       /* Now that we've connected to the dbus socket we know that / is populated.
        * We were able to connect to dbus with an absolute path, but to be certain
@@ -526,12 +528,12 @@ main (int   argc,
            */
           g_warning ("Initiating 10 minute shutdown timer due to error: %s", error->message);
           timeout_id = g_timeout_add_seconds (10 * 60, payg_sync_and_poweroff, NULL);
-          while (TRUE)
-            g_main_context_iteration (NULL, TRUE);
+          ret = NO_PROVIDER_EXIT_CODE;
         }
       else if (g_error_matches (error, GSS_SERVICE_ERROR, GSS_SERVICE_ERROR_SIGNALLED))
         {
           /* The service received SIGTERM or SIGINT */
+          timeout_id = g_idle_add (payg_sync_and_poweroff, NULL);
           ret = FATAL_SIGNAL_EXIT_CODE;
         }
       else
@@ -549,9 +551,14 @@ main (int   argc,
 
   allow_writing_to_boot_partition (FALSE);
 
+  /* If payg_sync_and_poweroff is scheduled, spin the mainloop until it runs
+   * and terminates the daemon.
+   */
+  while (timeout_id)
+    g_main_context_iteration (NULL, TRUE);
+
   if (ret == 0 && watchdog_id > 0)
     {
-      /* Continue to ping the watchdog indefinitely */
       g_message ("Entering watchdog-ping-only mode");
       while (TRUE)
         g_main_context_iteration (NULL, TRUE);
--- a/eos-paygd/main.c
+++ b/eos-paygd/main.c
@@ -531,9 +531,8 @@ main (int   argc,
         }
       else if (g_error_matches (error, GSS_SERVICE_ERROR, GSS_SERVICE_ERROR_SIGNALLED))
         {
-          g_warning ("EpgService exited with GSS_SERVICE_ERROR_SIGNALLED");
-          raise (gss_service_get_exit_signal (GSS_SERVICE (service)));
-          ret = FATAL_SIGNAL_EXIT_CODE; /* should not be reached, just in case the signal is caught */
+          /* The service received SIGTERM or SIGINT */
+          ret = FATAL_SIGNAL_EXIT_CODE;
         }
       else
         {
--- a/eos-paygd/main.c
+++ b/eos-paygd/main.c
@@ -281,7 +281,7 @@ main (int   argc,
           g_warning ("Unable to access EFI variables, shutting down in %d minutes",
                      TIMEOUT_POWEROFF_ON_ERROR_MINUTES);
           g_timeout_add_seconds (TIMEOUT_POWEROFF_ON_ERROR_MINUTES * 60,
-                                 payg_sync_and_poweroff, NULL);
+                                 payg_system_poweroff, NULL);
         }
 
       payg_set_debug_env_vars ();
@@ -318,7 +318,7 @@ main (int   argc,
           g_warning ("Security level regressed, shutting down in %d minutes",
                      TIMEOUT_POWEROFF_ON_ERROR_MINUTES);
           g_timeout_add_seconds (TIMEOUT_POWEROFF_ON_ERROR_MINUTES * 60,
-                                 payg_sync_and_poweroff, NULL);
+                                 payg_system_poweroff, NULL);
         }
 
       /* Setup RTC updater before the root pivot */
@@ -327,7 +327,7 @@ main (int   argc,
           g_warning ("RTC failure, shutting down in %d minutes",
                      TIMEOUT_POWEROFF_ON_ERROR_MINUTES);
           g_timeout_add_seconds (TIMEOUT_POWEROFF_ON_ERROR_MINUTES * 60,
-                                 payg_sync_and_poweroff, NULL);
+                                 payg_system_poweroff, NULL);
         }
     }
   else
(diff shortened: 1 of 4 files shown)


eos-paygd/main.c
```diff
@@ -532,7 +532,6 @@ main (int   argc,
       else if (g_error_matches (error, GSS_SERVICE_ERROR, GSS_SERVICE_ERROR_SIGNALLED))
         {
           g_warning ("EpgService exited with GSS_SERVICE_ERROR_SIGNALLED");
-          raise (gss_service_get_exit_signal (GSS_SERVICE (service)));
```
wjt, line 535: This dates all the way back to the initial commit in 150a34491830d49ec1a766c1eb617389effb372e. @pwithnall do you recall the rationale here?

eos-paygd/main.c
```diff
@@ -549,9 +551,12 @@ main (int   argc,
 
   allow_writing_to_boot_partition (FALSE);
 
+  /* If payg_sync_and_poweroff is scheduled, spin the mainloop until it runs */
+  while (timeout_id)
+      g_main_context_iteration (NULL, TRUE);
```
wjt, line 561: The coding style for this project would either use a 2-space indent here, or have some `{` `}` on surrounding lines at that indent.

eos-paygd/main.c
```diff
@@ -549,9 +551,12 @@ main (int   argc,
 
   allow_writing_to_boot_partition (FALSE);
 
+  /* If payg_sync_and_poweroff is scheduled, spin the mainloop until it runs */
+  while (timeout_id)
```
wjt: This is an infinite loop. `timeout_id` is local to this function, nothing will clear it even after the timeout or idle fires.

eos-paygd/main.c
```diff
@@ -552,7 +552,7 @@ main (int   argc,
   allow_writing_to_boot_partition (FALSE);
 
   /* If payg_sync_and_poweroff is scheduled, spin the mainloop until it runs */
-  while (timeout_id)
+  while (g_main_context_find_source_by_id (NULL, timeout_id))
```
wjt, line 560: OK, so this change corrects the infinite loop from the earlier patch.

eos-paygd/main.c
```diff
@@ -523,11 +527,12 @@ main (int   argc,
...
-          /* This could mean the PAYG data has been erased; force a poweroff
-           * after 10 minutes. See https://phabricator.endlessm.com/T27581
+          /* This could mean the PAYG data has been erased; force a poweroff.
+           * See https://phabricator.endlessm.com/T27581
            */
-          g_warning ("Initiating 10 minute shutdown timer due to error: %s", error->message);
-          timeout_id = g_timeout_add_seconds (10 * 60, payg_sync_and_poweroff, NULL);
+          g_warning ("Provider failure, shutting down in %d minutes: %s",
+                     POWEROFF_TIMEOUT_ERROR, error->message);
+          timeout_id = g_timeout_add_seconds (POWEROFF_TIMEOUT_ERROR * 60, payg_sync_and_poweroff, NULL);
```
wjt, line 535: What is the rationale for increasing this from 10 to 20? (I think it's probably OK.)

libeos-payg/util.c
```diff
@@ -34,20 +34,18 @@ static gboolean payg_legacy_mode = FALSE;
 __attribute__ ((__noreturn__)) gboolean
 payg_sync_and_poweroff (gpointer user_data)
```
wjt, line 35: This function is now misleadingly named.

wjt (changes requested):

eos-paygd/main.c
jprvita: I thought it was always a 4-space on lines containing code, and 2-space only before / after context delimiters (`{` `}`), but looking at [the spec](https://www.gnu.org/prep/standards/standards.html) you are actually right (I'm more of a K&R guy myself).

I'll fix this.

eos-paygd/main.c
jprvita: Nothing will clear it, but it is only infinite until `payg_sync_and_poweroff` is executed since at this commit it will call `exit`. I should have made it clear in the comment above, but as you saw it changes later, so it shouldn't matter too much.

The reason I have split the changes that follow as a separate commit was because I wasn't too happy with them, and expected them to change or be dropped during CR.

libeos-payg/util.c
jprvita: It still calls poweroff, but yeah, I agree it could be better named.

eos-paygd/main.c
pwithnall: `GSS_SERVICE_ERROR_SIGNALLED` is returned if a signal (by default, `SIGINT` or `SIGTERM`) is caught by the service. Rather than have the signal handlers try and clean up and exit, it seemed cleaner to just use that as a trigger for exiting the `gss_service_run()` call (which runs the main loop), and have the caller (i.e. this code) work out what it wants to do with the signal. Re-`raise()`ing it seemed like the correct thing to do in the absence of other requirements.

The code may want to do some process-wide cleanup or state saving before re-`raise()`ing the signal.

Essentially, the change proposed here means that the process will now swallow `SIGINT` and `SIGTERM` and, rather than exiting via `raise()`, it will warn and return `FATAL_SIGNAL_EXIT_CODE` from `main()`.

eos-paygd/main.c
jprvita: Just making all error paths have the same timeout length. Maybe this was intentionally shorter before because it was easier to damage the provider data storage? Now the storage is protected by the LSM (even when using the eMMC backend, to some extent), so I think it is fine to have the same error timeout.

eos-paygd/main.c
jprvita: Changed my mind slightly here while addressing other points. It does make sense to re-raise the signal for exiting so systemd has an accurate picture of what happened, but just at a different place now that we want to request a poweroff when that happen.

jprvita:
I have now addressed all the points here, plus something else discussed on the ticket. But I am changing this to a draft for now (feedback still welcome) and looking into how we handle hard shutdowns first, as it may influence this PR.

eos-paygd/main.c
pwithnall: :+1: :+1:

eos-paygd/main.c
```diff
@@ -35,11 +35,14 @@
 #include <sys/un.h>
 #include <libeos-payg/efi.h>
 
+#define POWEROFF_TIMEOUT_ERROR 20
```
pwithnall, line 38: Perhaps call it `POWEROFF_TIMEOUT_ON_ERROR_MINUTES`? Suffixing with `_ERROR` makes it look like this is an error or exit code. Suffixing with `_MINUTES` makes it clearer what the units are (as time is equally often measured in seconds or microseconds).

libeos-payg/manager-service.c
```diff
@@ -28,6 +28,7 @@
 #include <libeos-payg/manager-service.h>
 #include <libeos-payg/util.h>
 
+#define POWEROFF_TIMEOUT_NO_CREDIT 10
```
pwithnall, line 31: Perhaps `POWEROFF_TIMEOUT_NO_CREDIT_MINUTES`?

libeos-payg/util.c
```diff
@@ -31,24 +31,28 @@ static gboolean payg_legacy_mode = FALSE;
 /* Force a poweroff in situations where we are not able to enforce PAYG. This
  * is intended to be used as a GSourceFunc, e.g. with g_timeout_add_seconds()
```
pwithnall, line 32: Maybe expand the documentation to mention that if `user_data` is supplied, it must be a pointer to the source ID to be cleared.

libeos-payg/util.c
```diff
@@ -31,24 +31,28 @@ static gboolean payg_legacy_mode = FALSE;
...
-   * Exit now so the watchdog timer will make sure we get our shutdown
-   * eventually.
-   */
-  exit (EXIT_FAILURE);
+
+  g_message ("Requesting an orderly system shudown");
+  ret = system ("systemctl poweroff");
+  g_debug ("systemctl returned %d", ret);
+
+  /* We have requested an orderly system shutdown. If that request failed of if
```
pwithnall, line 49:
```suggestion
  /* We have requested an orderly system shutdown. If that request failed or if
```

libeos-payg/util.c
```diff
@@ -31,24 +31,28 @@ static gboolean payg_legacy_mode = FALSE;
...
+   * a shutdown was already in progress 'systemctl' will return a failure
+   * status, and we can't distinguish between these two scenarios. So let's
```
pwithnall, line 51: Don’t suppose we could submit a change to `systemctl` to get it to return a different error code if shutdown is already in progress?

Or perhaps we can get the same effect by calling `poweroff` rather than `systemctl poweroff`?

eos-paygd/main.c
jprvita: Interesting, I usually expect error codes to be prefixed with `ERROR` instead of suffixed. I don't care too much, so if it makes it easier for others to read I'll take this. But I actually went with `TIMEOUT_POWEROFF_NO_CREDIT_MINUTES`, which seems more organized.

libeos-payg/manager-service.c
jprvita: ditto, `TIMEOUT_POWEROFF_NO_CREDIT_MINUTES`.

libeos-payg/util.c
jprvita: Good point.

libeos-payg/util.c
jprvita: Good catch!

libeos-payg/util.c
jprvita: On modern systems, `poweroff` is simply a compat entry point for `systemctl poweroff` and has the same return values. We could try that, but I am not convinced that it makes sense to return a success status if the tool failed to request a shutdown because one is already in progress (IDK if systemd / logind even return this info to systemctl).

If we really want to continue forcing a poweroff, we could try to look at `systemctl is-system-running` (which I believe is the same as the `SystemState` property in systemd's D-Bus API) or whether shutdown.target is active, but I'm not sure it is worth since ultimately none of those bits are protected by our security measures.

I have tried to improve the commit message a bit to make this clearer.

pwithnall (approved):
LGTM from my point of view now

PR #96, opened 2022-02-11
paygd: Log code entering attemps and results

jprvita:
https://phabricator.endlessm.com/T32734

### DIFF
--- a/libeos-payg/manager-service.c
+++ b/libeos-payg/manager-service.c
@@ -999,12 +999,19 @@ epg_manager_service_manager_add_code (EpgManagerService     *self,
   const gchar *code_str;
   g_variant_get (parameters, "(&s)", &code_str);
 
+  g_message ("Trying to enter code %s", code_str);
   epg_provider_add_code (self->provider, code_str, &time_added, &local_error);
 
   if (local_error != NULL)
-    g_dbus_method_invocation_return_gerror (invocation, local_error);
+    {
+      g_message ("Failed to enter code %s: %s", code_str, local_error->message);
+      g_dbus_method_invocation_return_gerror (invocation, local_error);
+    }
   else
-    g_dbus_method_invocation_return_value (invocation, g_variant_new ("(x)", time_added));
+    {
+      g_message ("Added %" G_GINT64_FORMAT " units of credit", time_added);
+      g_dbus_method_invocation_return_value (invocation, g_variant_new ("(x)", time_added));
+    }
 }
 
 static void


libeos-payg/manager-service.c
```diff
@@ -993,12 +993,19 @@ epg_manager_service_manager_add_code (EpgManagerService     *self,
   const gchar *code_str;
   g_variant_get (parameters, "(&s)", &code_str);
 
+  g_message ("Trying to enter code %s", code_str);
   epg_provider_add_code (self->provider, code_str, &time_added, &local_error);
 
   if (local_error != NULL)
-    g_dbus_method_invocation_return_gerror (invocation, local_error);
+    {
+      g_warning ("Failed to enter code: %s", local_error->message);
```
wjt, line 1001: Is this warning-worthy?

```suggestion
      g_warning ("Failed to enter code %s: %s", code_str, local_error->message);
```

(I know it's redundant with the line above.)

jprvita: > Is this warning-worthy?

I think so, as in this is not the expected side of the branch. Another way of looking at is that it is just a warning, not a critical warning or an error. But I'm happy to reduce if you think it doesn't make sense.

I'll squash the suggested change to the original commit and force-push.

libeos-payg/manager-service.c
```diff
@@ -999,12 +999,19 @@ epg_manager_service_manager_add_code (EpgManagerService     *self,
   const gchar *code_str;
   g_variant_get (parameters, "(&s)", &code_str);
 
+  g_message ("Trying to enter code %s", code_str);
   epg_provider_add_code (self->provider, code_str, &time_added, &local_error);
 
   if (local_error != NULL)
-    g_dbus_method_invocation_return_gerror (invocation, local_error);
+    {
+      g_warning ("Failed to enter code %s: %s", code_str, local_error->message);
```
wjt, line 1007: I still don't think this is warning-worthy; I tend to think of a warning as a "soft assertion failure" rather than "a string was passed to a D-Bus method that I didn't like". GLib has a runtime flag to make warnings fatal to aid debugging, [`G_DEBUG=fatal-warnings`](https://docs.gtk.org/glib/running.html) and some projects like to set this on their test suite; having a warning here would prevent that.

wjt (changes requested):

libeos-payg/manager-service.c
wjt: (I should have been clearer in my original comment.)

wjt: Oh also, typo in commit message: `s/attemps/attempts/`

jprvita: Addressed both!

wjt (approved):
LGTM

PR #98, opened 2022-03-29
Revive manual-provision provisioning tarballs

jprvita:
This brings back commit 64a18485b2 "Add factory test script for Endless
backend" / reverts commit 0f6771b18c "manual-provision: Remove
provisioning tarballs", as we plan to use this phase-1 backend for a
field test.
    
https://phabricator.endlessm.com/T33306
https://phabricator.endlessm.com/T31277
https://phabricator.endlessm.com/T23967


manual-provision-payg/config.json
```diff
@@ -1,5 +1,4 @@
 {
-"InstructionsLine1": "To get a new unlock code send the amount required to Caletre, with the account number as $device_id.",
-"InstructionsLine2": "You will then receive an unlock code by SMS. Use this code to unlock time on your computer."
-
+"InstructionsLine1": "To get a new unlock keycode please contact your vendor with the account number as $device_id.",
+"InstructionsLine2": "You will then receive a keycode by SMS to unlock and add usage time to your computer."
```
wjt, line 3: We'll probably need to translate these, depending on the location of the field trials :)

wjt (commented):
Broadly LGTM

manual-provision-payg/hide_provenance.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+sha256sum "$1" | sed -e s\\"$1"\\`basename "$1"`\\g
```
wjt, line 3: This seems rather obscure; how about:

```bash
#!/bin/bash
set -e
cd "$(dirname $1)"
sha256sum "$(basename $1)"
```

wjt: Also perhaps `sha256sum_without_dirname.sh` as a better filename?

wjt (changes requested):

manual-provision-payg/hide_provenance.sh
wjt: Hmm, OK.

jprvita:
I've force-pushed after making sure every commit builds and passes unit tests. Since we have the factory tarballs checked-out in the repository and a test that compares the tarball produced in the build with the checked-out one, that means every time we make a change to the provisioning scripts we need to build, copy the tarballs back to the source dir and add them to the same commit :/

Maybe we should arrange for these tarballs to be published through the PAYG-provisioning-tarballs CI job, or we just note it and move on. In any case, we should at least manually publish them together with the phase 4 ones, so they are all in the same place and we get a nice direct link.

wjt:
Yes, I regret this model of checking in the generated files... Should have done something with CI instead, in hindsight it wouldn't have taken any longer than the time wasted regenerating stuff and manually committing it.

PR #99, opened 2022-03-30
manager: Use the device ID as the PAYG account ID

jprvita:
When using our simplistic default backend the account ID is left blank.
We do, however, generate a device ID during provisioning, from the first
8 digits of the machine ID at provisioning time, which is used by the
vendor to identify the account associated with the device.

Let's save this ID to the disk during provisioning so paygd can load it
and display it in the UI as the account ID.

https://phabricator.endlessm.com/T33322


jprvita:
Created as a draft  as we should merge https://github.com/endlessm/eos-payg/pull/98 first.

libeos-payg/manager.c
```diff
@@ -1329,6 +1337,14 @@ file_load_cb (GObject      *source_object,
       self->key_bytes = g_bytes_new_take (g_steal_pointer (&data), data_len);
       data_len = 0;
     }
+  else if (g_file_equal (file, self->account_id_file))
+    {
+      if (data != NULL)
+        {
+          g_free ((gchar *) self->account_id);
```
wjt, line 1344: The field shouldn't be declared `const`. Remove that qualifier and the cast won't be needed.

libeos-payg/manager.c
```diff
@@ -1329,6 +1337,14 @@ file_load_cb (GObject      *source_object,
...
+          self->account_id = g_strndup (data, data_len);
```
wjt, line 1345: You don't need to copy it.

```
self->account_id = g_steal_pointer (&data);
```

libeos-payg/manager.c
wjt: The same cast in `finalize` can be removed too.

libeos-payg/manager.c
```diff
@@ -335,6 +336,8 @@ epg_manager_constructed (GObject *object)
     self->clock = EPG_CLOCK (epg_real_clock_new ());
 
   self->account_id = g_strdup ("");
+  if (self->account_id_file == NULL)
+    self->account_id_file = g_file_new_for_path (PREFIX "/local/share/eos-payg/device-id");
```
wjt, line 340: Since the account ID is derived from the machine ID – why not read the first N characters of `/etc/machine-id`?

wjt (changes requested):

libeos-payg/manager.c
jprvita: Ok, I was trying to keep the changes to a minimum, but I agree that makes sense.

libeos-payg/manager.c
jprvita: Because the machine-id could have changed since (although we would hope it did not). Also, it seems better to keep it together with other eos-paygd's data.

libeos-payg/manager.c
```diff
@@ -335,6 +336,8 @@ epg_manager_constructed (GObject *object)
     self->clock = EPG_CLOCK (epg_real_clock_new ());
 
   self->account_id = g_strdup ("");
+  if (self->account_id_file == NULL)
```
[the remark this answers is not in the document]
jprvita: Great! I ran out of time and energy to even think about unit tests yesterday, but today is a new day :sun_with_face:

libeos-payg/manager.c
jprvita: hum, actually, I think this is correct. `const <type> *` will mean the contents of the array are immutable, which is what we want (although I don't think it matters too much in this case either way). Except if this is any different since the type here is a gchar instead of regular char, but I bet not.

I'll keep it as-is for now.

jprvita:
I have addressed earlier feedback, added unit tests (which was more work than expected, since I had to add a new construct-only property to EpgManager) and made sure every commit always passes unit tests.

I've also decided to rename the file from device-id to account-id, since in the context of eos-paygd that is always referred to as account-id (we could also rename it in the provisioning script and old documentation; it does not seem worth at this point, but maybe when we write new docs for this pilot we take it on).

libeos-payg/tests/manager.c
```diff
@@ -67,6 +72,7 @@ setup (Fixture *fixture,
   fixture->used_codes_path = g_build_filename (fixture->tmp_path, "used-codes", NULL);
 
   fixture->key = g_bytes_new_static (KEY, sizeof (KEY) - 1);
+  fixture->account_id = g_strdup (ACCOUNT_ID);
```
dbnicholson, line 75: This doesn't seem to be used anywhere.

libeos-payg/manager.c
dbnicholson: But you're not treating the pointed to value immutably here since you're freeing the data and the location this is pointing to in the heap is something else entirely. It's not the end of the world, but it seems pretty clearly wrong. It's trying to express that the value can't be changed, but `const` just doesn't capture those semantics in this use case.

libeos-payg/tests/manager.c
jprvita: Good catch!

libeos-payg/manager.c
jprvita: Ok, I think my confusion here is that I never realized that freeing a pointer is considered modifying the memory area pointed by it. And the assignment is just changing the address being pointed to, not the contents of the memory area.

Anyway, I've fixed it now.

dbnicholson (approved):

PR #100, opened 2022-04-01
Create CodeLength D-Bus property

jprvita:
Create a new CodeUIString read-only D-Bus property that will be used by
providers to inform the UI on how their unlock codes should be referred
to.

This also overrides its value in EpgManager with 8.

https://phabricator.endlessm.com/T33325

### DIFF
--- a/libeos-payg/manager-service.c
+++ b/libeos-payg/manager-service.c
@@ -993,7 +993,7 @@ epg_manager_service_manager_get_code_length (EpgManagerService     *self,
                                              const gchar           *property_name,
                                              GDBusMethodInvocation *invocation)
 {
-  return g_variant_new_uint32 (0);
+  return g_variant_new_uint32 (epg_provider_get_code_length (self->provider));
 }
 
 static GVariant *
--- a/libeos-payg/manager-service.c
+++ b/libeos-payg/manager-service.c
@@ -130,6 +130,14 @@ static GVariant *epg_manager_service_manager_get_code_format_suffix (EpgManagerS
                                                                      const gchar           *interface_name,
                                                                      const gchar           *property_name,
                                                                      GDBusMethodInvocation *invocation);
+
+static GVariant *epg_manager_service_manager_get_code_length (EpgManagerService     *self,
+                                                              GDBusConnection       *connection,
+                                                              const gchar           *sender,
+                                                              const gchar           *interface_name,
+                                                              const gchar           *property_name,
+                                                              GDBusMethodInvocation *invocation);
+
 static GVariant *epg_manager_service_manager_get_account_id (EpgManagerService     *self,
                                                              GDBusConnection       *connection,
                                                              const gchar           *sender,
@@ -676,6 +684,8 @@ manager_properties[] =
       epg_manager_service_manager_get_code_format_prefix, NULL  /* read-only */ },
     { "com.endlessm.Payg1", "CodeFormatSuffix", "code-format-suffix",
       epg_manager_service_manager_get_code_format_suffix, NULL  /* read-only */ },
+    { "com.endlessm.Payg1", "CodeLength", "code-ui-string",
+      epg_manager_service_manager_get_code_length, NULL  /* read-only */ },
     { "com.endlessm.Payg1", "AccountID", "account-id",
       epg_manager_service_manager_get_account_id, NULL  /* read-only */ },
   };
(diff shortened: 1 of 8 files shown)


libeos-payg/manager-service.c
```diff
@@ -676,6 +684,8 @@ manager_properties[] =
       epg_manager_service_manager_get_code_format_prefix, NULL  /* read-only */ },
     { "com.endlessm.Payg1", "CodeFormatSuffix", "code-format-suffix",
       epg_manager_service_manager_get_code_format_suffix, NULL  /* read-only */ },
+    { "com.endlessm.Payg1", "CodeUIString", "code-ui-string",
```
pwithnall, line 687: I don’t think this approach works for localisation, for two reasons:
 1. Translation is not homomorphic: `_("Enter %s below").format(_("8-digit keycode"))` will not always give the same results as `_("Enter %s below".format("8-digit keycode"))` in all languages
 2. Runtime translation has to happen in the same module as the strings are defined, otherwise they won’t be extracted into the right gettext catalogues at compile time (which precludes translating after substituting `CodeUIString` into a format placeholder)
 3. `eos-paygd` might be running in a different locale from the user’s UI process — this is not a theoretical problem; users do run systems where the greeter runs in a different language from one or more users’ sessions

Adding all those up, I think there are three options here:
 1. Avoid anything to do with translatable strings in the API for `Payg1`, which probably means adding a `CodeFormat` D-Bus property which is defined as an enum `["phase1", "angaza", "paygee"]`. The client would then interpret this to produce a translated prompt string as appropriate. If the client didn’t recognise the format enum, it’d fall back to prompting for a generic ‘keycode’.
 2. Alternatively, add a `CodeLength` D-Bus property which returns a tuple `(7, 19)` defining the minimum and maximum lengths of the keycode. Zero would indicate an open bound. If they’re both zero the format would be undefined and the client should just say ‘keycode’. This approach would avoid having to update the client if a new keycode format was added in future (as long as the addition is a sequence of digits of some known length).
 3. Or, for full flexibility in defining the prompt string, add a `GetCodePrompt(s locale)` function to the D-Bus interface which, when called with the locale of the client process, returns a translated string to display. This would be the full prompt string, “Enter a 7–19 digit keycode to unlock your computer”. This approach is more complex but allows for the prompt string to be provided by the daemon rather than the client. The big downside of that is that the prompt string then might not fit with the client’s UI particularly well.

My suggestion would be to go with approach #1. It gives a reasonable amount of flexibility, and avoids all the issues of sending translatable strings over D-Bus. Its downside is that the client code would need to be updated to handle new keycode formats if they were added in future.

jprvita: I went with a simplified version of approach 2, using just an integer for a fixed code length (and 0 meaning not defined), as that is really all we need to support right now. We can extend it in the future if we ever need to.

libeos-payg/provider.c
```diff
@@ -154,6 +154,24 @@ epg_provider_default_init (EpgProviderInterface *iface)
                            G_PARAM_READABLE | G_PARAM_STATIC_STRINGS);
   g_object_interface_install_property (iface, pspec);
 
+  /**
+   * EpgProvider:code-length:
+   *
+   * The length of an unlock code.
```
pwithnall, line 160:
```suggestion
   * The length of an unlock code, in Unicode characters.
```

Should probably be explicit about whether the length is in terms of characters, or bytes. Characters would be more flexible for accepting emojis in future (for example).

libeos-payg/provider.c
```diff
@@ -154,6 +154,24 @@ epg_provider_default_init (EpgProviderInterface *iface)
...
+   *
+   * This property is constant once the provider is initialized.
+   *
+   * A value of 0 means the length was not specified by the provider.
+   *
+   * Since: 0.2.4
+   */
+  pspec =
+      g_param_spec_uint ("code-length", "Code Length",
+                         "The length of an unlock code.",
```
pwithnall, line 170:
```suggestion
                         "The length of an unlock code, in Unicode characters.",
```

libeos-payg/provider.c
```diff
@@ -506,6 +524,27 @@ epg_provider_get_code_format_suffix (EpgProvider *self)
...
 }
 
+/**
+ * epg_provider_get_code_length:
+ * @self: a #EpgProvider
+ *
+ * Get the value of #EpgProvider:code-length
+ *
+ * Returns: the value to be used by the UI when referring to the unlock code
+ * length. A value of 0 means the UI should not mention the unlock code length
```
pwithnall, line 534:
```suggestion
 * length. in Unicode characters. A value of 0 means the UI should not mention the unlock code length
```

pwithnall (approved):
LGTM. One suggestion about clarifying the units the length is in

jprvita:
Squashed the suggested changes into the original commit, fixing a typo (`.`->`,`) and line breaks.

pwithnall (approved):
LGTM :D

PR #101, opened 2022-04-08
manual-provision: Create per-machine dir for artifacts

jprvita:
Instead of always saving the artifacts from the provisioning script to
the same location, create a per-machine directory in the provisioning
USB device, named with the machine's device id (which is derived from
/etc/machine-id).

https://phabricator.endlessm.com/T33349


dbnicholson (approved):
I'm not familiar with this tool but it all seems totally reasonable!

PR #102, opened 2022-04-11
provision-phase-1: Build and publish tarballs in Jenkins

jprvita:
https://phabricator.endlessm.com/T33373


dbnicholson (approved):
That all looks good. You probably already knew this, but I checked again and jenkins is allowed to write to that part of the S3 bucket. The `--max-age=300` (5 minutes) is probably too short since I presume these will change very rarely, but it doesn't really hurt.

provision-phase-1/Jenkinsfile
```diff
@@ -0,0 +1,34 @@
+pipeline {
+    agent {
+        dockerfile {
+            filename 'Dockerfile.build'
+
+            // Install awscli for uploading the built tarballs to S3.
+            additionalBuildArgs "--build-arg=packages=awscli"
```
dbnicholson, line 7: Not sure it really matters, but the jenkins job has a `BRANCH` parameter for the OS and the `Dockerfile` supports a `branch` build arg, but it's not passed through here.

jprvita: In theory we should be building in an OS branch that matches the code branch. In practice I don't think it really matters, since all that is being called in the build is `tar` and `sha256sum`, and PAYG repos have dev and stable branches almost always flush with one another.

If we change it here we should also change it for the other jobs in `payg.yaml`

jprvita:
> That all looks good. You probably already knew this, but I checked again and jenkins is allowed to write to that part of the S3 bucket. The `--max-age=300` (5 minutes) is probably too short since I presume these will change very rarely, but it doesn't really hurt.

I'm happy to go with a different value, but we should change it for the other PAYG tarballs that get published to S3 as well (they are all expected to change very rarely).

provision-phase-1/Jenkinsfile
dbnicholson: I suggest `additionalBuildArgs "--build-arg=packages=awscli --build-arg=branch=${params.BRANCH}"`. Then the OS branch will match the git branch and the build deps are more likely to match.

dbnicholson:
> > That all looks good. You probably already knew this, but I checked again and jenkins is allowed to write to that part of the S3 bucket. The `--max-age=300` (5 minutes) is probably too short since I presume these will change very rarely, but it doesn't really hurt.
> 
> I'm happy to go with a different value, but we should change it for the other PAYG tarballs that get published to S3 as well (they are all expected to change very rarely).

Let's just leave it for now.

provision-phase-1/Jenkinsfile
dbnicholson: I think it's not running `tar` that's an issue, but rather running `meson` since it will try find all the dependencies for the whole project.

jprvita: Yep, I know. That was why I also mentioned that the branches are almost always flush, so I don't think we'll hit problems with dependencies.

In any case, I agree that passing the branch is the right thing to do, sorry for going on a unnecessary tangent. I'll update this PR.

jprvita:
Updated to pass the branch parameter to docker when building the container.

dbnicholson (approved):

PR #103, opened 2022-04-12
provision-phase-1: Update git submodules before building

jprvita:
Even if we are not actually building any submodules when generating the
tarballs, we need the tree to be present during meson configuration.

https://phabricator.endlessm.com/T33373

### DIFF
--- a/provision-phase-1/Jenkinsfile
+++ b/provision-phase-1/Jenkinsfile
@@ -11,6 +11,7 @@ pipeline {
     stages {
         stage('Build') {
             steps {
+                sh 'git submodule update'
                 sh 'meson . build'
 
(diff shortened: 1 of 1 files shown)


dbnicholson:
This is fine, but I think it would be a little more natural to have jenkins do this since it's doing the checkout. See endlessm/eos-build#1339.

jprvita:
> This is fine, but I think it would be a little more natural to have jenkins do this since it's doing the checkout. See [endlessm/eos-build#1339](https://github.com/endlessm/eos-build/pull/1339).

Sounds good to me! I merged [endlessm/eos-build#1339](https://github.com/endlessm/eos-build/pull/1339) now so I'm closing this one.

PR #104, opened 2022-04-20
paygd: Write logs to a separate file

jprvita:
In systems with fragile storage (ex., SSD or eMMC) the journal is only
kept in RAM as does not persist over reboot. In such systems we still
want logs from eos-paygd to persist, without having to write all system
logs to persistent storage.

The solution for that is to create a separate log file for eos-paygd, in
addition to forwarding logs to the journal.

https://phabricator.endlessm.com/T33220

### DIFF
--- a/eos-paygd/main.c
+++ b/eos-paygd/main.c
@@ -35,7 +35,61 @@
 #include <sys/un.h>
 #include <libeos-payg/efi.h>
 
+#define LOG_FILE_NAME "/var/log/eos-payg.log"
+
 #define TIMEOUT_POWEROFF_ON_ERROR_MINUTES 20
+
+/* we can't use g_log_writer_default_would_drop until glib 2.68, which will be
+ * available on EOS 5.0+, so we are copying a simplified version of its
+ * implementation here.
+ */
+#define DEFAULT_LEVELS (G_LOG_LEVEL_ERROR | G_LOG_LEVEL_CRITICAL | G_LOG_LEVEL_WARNING | G_LOG_LEVEL_MESSAGE)
+#define INFO_LEVELS (G_LOG_LEVEL_INFO | G_LOG_LEVEL_DEBUG)
+static gboolean
+should_drop_message (GLogLevelFlags log_level)
+{
+  /* Disable debug message output unless specified in G_MESSAGES_DEBUG. */
+  if (!(log_level & DEFAULT_LEVELS) && !(log_level >> G_LOG_LEVEL_USER_SHIFT))
+    {
+      const gchar *domains = g_getenv ("G_MESSAGES_DEBUG");
+
+      if ((log_level & INFO_LEVELS) == 0 ||
+          domains == NULL)
+        return TRUE;
+
+      if (strcmp (domains, "all") != 0)
+        return TRUE;
+    }
+
+  return FALSE;
+}
+
+/* Custom log writer function that saves messages to a separate file before
+ * forwarding them to the default writer function.
+ */
+static GLogWriterOutput
+log_writer (GLogLevelFlags log_level,
+            const GLogField* fields,
+            gsize n_fields,
+            gpointer user_data)
+{
+  g_autofree gchar *out = NULL;
+  GIOChannel *log_io = (GIOChannel *) user_data;
+
+  g_return_val_if_fail (fields != NULL, G_LOG_WRITER_UNHANDLED);
+  g_return_val_if_fail (n_fields > 0, G_LOG_WRITER_UNHANDLED);
+
+  if (should_drop_message (log_level))
+    return G_LOG_WRITER_HANDLED;
+
+  out = g_log_writer_format_fields (log_level, fields, n_fields, FALSE);
+  g_io_channel_write_chars (log_io, out, -1, NULL, NULL);
+  g_io_channel_write_chars (log_io, "\n", 1, NULL, NULL);
+  g_io_channel_flush (log_io, NULL);
+
+  /* now pass the message to the the default writer function as well */
+  return g_log_writer_default (log_level, fields, n_fields, user_data);
+}
 
 static int watchdog_fd = -1;
 
@@ -231,6 +285,7 @@ main (int   argc,
   g_autoptr(GError) error = NULL;
   g_autoptr(EpgService) service = NULL;
   g_autoptr(GFile) state_dir = NULL;
+  GIOChannel *log_io = NULL;
   int ret = EXIT_SUCCESS, sd_notify_ret, system_ret, exit_signal = 0;
   int lsm_fd;
   guint timeout_id = 0, watchdog_id = 0;
@@ -487,6 +542,19 @@ main (int   argc,
       eospayg_efi_root_pivot ();
     }
 
+  /* Try to log to a file in /var, in addition to the journal, so we have
+   * persistent logs even on systems with fragile storage.
+   */
+  log_io = g_io_channel_new_file (LOG_FILE_NAME, "a", &error);
+  if (!log_io)
+    {
+      g_warning ("Failed to open %s: %s", LOG_FILE_NAME, error->message);
+    }
+  else
+    {
+      g_log_set_writer_func (log_writer, log_io, NULL);
+    }
+
   /* Technically this existence check is racy but no other process should be
    * accessing this path
    */
@@ -568,6 +636,9 @@ main (int   argc,
       while (TRUE)
         g_main_context_iteration (NULL, TRUE);
     }
+
+  if (log_io)
+    g_io_channel_unref (log_io);
 
   return ret;
 }
--- a/eos-paygd/main.c
+++ b/eos-paygd/main.c
@@ -44,23 +44,10 @@
 static GIOChannel *
 open_log_file (void)
 {
-  int system_ret;
   const gchar *log_file_name = NULL;
   g_autoptr(GFile) log_file = NULL;
   g_autoptr(GDateTime) now = NULL;
   g_autofree gchar *tstamp = NULL;
-  g_autoptr(GError) error = NULL;
-  GIOChannel *log_io = NULL;
-
-  /* Use /bin/mkdir instead of mkdir() to ensure the mode is unaffected by
-   * the process's umask.
-   */
-  system_ret = system ("/bin/mkdir -p --mode=755 " LOGFILE_DIRNAME);
-  if (system_ret == -1 || !WIFEXITED (system_ret) || WEXITSTATUS (system_ret) != 0)
-    {
-      g_warning ("mkdir of %s failed", LOGFILE_DIRNAME);
-      return NULL;
-    }
 
   /* Build log file name with a date stamp so we get one file per day,
    * ex., eos-paygd-20220418.log
@@ -71,11 +58,9 @@ open_log_file (void)
                                tstamp, ".", LOGFILE_EXT, NULL);
   log_file = g_file_new_for_path (log_file_name);
 
-  log_io = g_io_channel_new_file (log_file_name, "a", &error);
-  if (!log_io)
-    g_warning ("Failed to open %s: %s", log_file_name, error->message);
-
-  return log_io;
+  /* We can't log an error here if this fails, as it would cause infinite
+   * recursion */
+  return g_io_channel_new_file (log_file_name, "a", NULL);
 }
 
 /* we can't use g_log_writer_default_would_drop until glib 2.68, which will be
@@ -112,8 +97,7 @@ log_writer (GLogLevelFlags log_level,
             gsize n_fields,
             gpointer user_data)
 {
-  g_autofree gchar *out = NULL;
-  GIOChannel *log_io = (GIOChannel *) user_data;
+  GIOChannel *log_io = NULL;
 
   g_return_val_if_fail (fields != NULL, G_LOG_WRITER_UNHANDLED);
   g_return_val_if_fail (n_fields > 0, G_LOG_WRITER_UNHANDLED);
@@ -121,10 +105,15 @@ log_writer (GLogLevelFlags log_level,
   if (should_drop_message (log_level))
     return G_LOG_WRITER_HANDLED;
 
-  out = g_log_writer_format_fields (log_level, fields, n_fields, FALSE);
-  g_io_channel_write_chars (log_io, out, -1, NULL, NULL);
-  g_io_channel_write_chars (log_io, "\n", 1, NULL, NULL);
-  g_io_channel_flush (log_io, NULL);
+  log_io = open_log_file ();
+  if (log_io)
+    {
+      g_autofree gchar *out = NULL;
+      out = g_log_writer_format_fields (log_level, fields, n_fields, FALSE);
+      g_io_channel_write_chars (log_io, out, -1, NULL, NULL);
+      g_io_channel_write_chars (log_io, "\n", 1, NULL, NULL);
+      g_io_channel_unref (log_io); /* flush + close + free */
+    }
 
   /* now pass the message to the the default writer function as well */
   return g_log_writer_default (log_level, fields, n_fields, user_data);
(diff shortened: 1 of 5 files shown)


eos-paygd/main.c
```diff
@@ -240,6 +295,9 @@ main (int   argc,
   GOptionContext *context;
   gboolean enforcing_mode = TRUE;
 
+  log_io = g_io_channel_new_file (LOG_FILE_NAME, "a", NULL);
```
wjt: Is the real `/var` accessible when `eos-payg` starts up?

You are not handling failure from this function, indicated by it returning `NULL`. https://docs.gtk.org/glib/ctor.IOChannel.new_file.html

You could write:

```
  log_io = g_io_channel_new_file (LOG_FILE_NAME, "a", &error);
  if (log_io != NULL)
    {
      g_log_set_writer_func (log_writer, log_io, NULL);
    }
  else
    {
      g_warning ("Failed to open %s: %s", LOG_FILE_NAME, error->message);
      g_clear_error (&error);
    }
```

eos-paygd/main.c
```diff
@@ -268,6 +271,76 @@ test_and_update_securitylevel (void)
...
+  int system_ret;
+  g_autoptr(GFile) log_file = NULL;
+  g_autoptr(GFileInfo) log_file_info = NULL;
+  g_autoptr(GError) error = NULL;
+  const gchar *log_file_name = NULL;
+
+  /* Use /bin/mkdir instead of mkdir() to ensure the mode is unaffected by
+   * the process's umask.
+   */
+  system_ret = system ("/bin/mkdir -p --mode=755 " LOGFILE_DIRNAME);
```
wjt, line 295: Can we create the directory with `systemd-tmpfiles` instead?

wjt: Or, looks like we can use `LogsDirectory=eos-paygd` in the unit file to cause systemd to create it before launching eos-paygd.

eos-paygd/main.c
```diff
@@ -271,6 +272,14 @@ test_and_update_securitylevel (void)
   return TRUE;
 }
 
+static gint
+sort_func (gconstpointer p1, gconstpointer p2)
+{
+  const gchar *a = (const char *) p1;
+  const gchar *b = (const char *) p2;
+  return g_strcmp0 (a, b);
+}
```
wjt, line 281: You don't need this, you can just pass `g_strcmp0` as the compare_func.

wjt (changes requested):
I confess to not having read the cleanup logic in detail.

I wonder if we could instead use systemd-tmpfile to log files older than a certain age; and rather than having logic to rename old files, always name the log file after the current timestamp. This would be a lot less code!

eos-paygd/main.c
```diff
@@ -240,6 +295,9 @@ main (int   argc,
   GOptionContext *context;
   gboolean enforcing_mode = TRUE;
 
+  log_io = g_io_channel_new_file (LOG_FILE_NAME, "a", NULL);
+  g_log_set_writer_func (log_writer, log_io, NULL);
```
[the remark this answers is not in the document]
jprvita: Hum, I thought this would try to free `user_data` after each call to the callback. Are you sure the way you describe is how it works?

I supposed I could also close the io channel on the last line before returning from main as well.

eos-paygd/main.c
```diff
@@ -35,8 +35,62 @@
...
+ * forwarding them to the default writer function.
+ */
+static GLogWriterOutput
+log_writer (GLogLevelFlags log_level,
+            const GLogField* fields,
+            gsize n_fields,
+            gpointer user_data)
+{
+  g_autofree gchar *out = NULL;
+  GIOChannel *log_io = (GIOChannel *) user_data;
```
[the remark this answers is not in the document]
jprvita: Thanks!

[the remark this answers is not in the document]
wjt: Or maybe not, this is not a GObject!

eos-paygd/main.c
wjt: Cast it: `(GCompareFunc) g_strcmp0`

jprvita: :brain:

eos-paygd/main.c
jprvita: > Is the real `/var` accessible when `eos-payg` starts up?
> 

Hum, you are right, we need to wait until pivoting to the real root, which means we'll lose some startup messages (I). That is fine IMO, as these will still be in the journal and are either fatal errors or redundant debug info: nothing that would affect an analysis of credit consumption.

eos-paygd/main.c
jprvita: I don't think so, as `systemd-tmpfiles` would not have run before paygd starts, and `LogsDirectory=` would add a `RequiresMountsFor=/var` dependency which I believe would either delay paygd's startup or create an impossible transaction.

I'm using `mkdir` following how `/var/lib/eos-payg` is created a few lines below.

jprvita:
Addressed feedback provided in this PR, plus @wjt's suggestion to split logs as one file per day then use systemd-tmpfiles to cleanup files older then a certain age (I have chosen 90 days).

(I haven't been able to test the 90 days removal yet)

jprvita:
Ah, one minor annoyance that I noticed when testing the build changes:
```
tmpfilesdir = get_option('tmpfilesdir')
if tmpfilesdir == ''
  tmpfilesdir = systemd.get_pkgconfig_variable('tmpfilesdir')
endif
```
This ends up installing the tmpfile in the system's `/usr` instead of under the prefix. The same is true for the unit files. It does not matter unless you build and install eos-payg locally for some reason. It would be nice to fix if someone happens to know the right way from the top of their head, but it is not worth putting any actual effort into it IMO.

eos-paygd/meson.build
```diff
@@ -36,6 +36,10 @@ systemdsystemunitdir = get_option('systemdsystemunitdir')
 if systemdsystemunitdir == ''
   systemdsystemunitdir = systemd.get_pkgconfig_variable('systemdsystemunitdir')
 endif
+tmpfilesdir = get_option('tmpfilesdir')
+if tmpfilesdir == ''
+  tmpfilesdir = systemd.get_pkgconfig_variable('tmpfilesdir')
+endif
```
wjt, line 42: Try this:

```suggestion
tmpfilesdir = get_option('tmpfilesdir')
if tmpfilesdir == ''
  tmpfilesdir = systemd.get_pkgconfig_variable('tmpfilesdir', define_variable: ['prefix', prefix])
endif
```

Found at https://gitlab.freedesktop.org/pipewire/pipewire/-/merge_requests/235/diffs

eos-paygd/main.c
```diff
@@ -35,10 +35,51 @@
...
+    {
+      g_warning ("mkdir of %s failed", LOGFILE_DIRNAME);
+      return FALSE;
+    }
+
+  /* Build log file name with a date stamp so we get one file per day,
+   * ex., eos-paygd-20220418.log
+   */
+  now = g_date_time_new_now_local ();
+  tstamp = g_date_time_format (now, "%Y%m%d");
```
wjt, line 57: Why not `%F`, the ISO-8601 date format (`%Y-%m-%d`)?

eos-paygd/main.c
```diff
@@ -583,11 +574,22 @@ main (int   argc,
...
     }
 
-  /* Try to log to a file in /var, in addition to the journal, so we have
-   * persistent logs even on systems with fragile storage.
+  /* Use /bin/mkdir instead of mkdir() to ensure the mode is unaffected by
+   * the process's umask.
    */
-  if (open_log_file (&log_io))
-    g_log_set_writer_func (log_writer, log_io, NULL);
+  system_ret = system ("/bin/mkdir -p --mode=755 " LOGFILE_DIRNAME);
```
wjt, line 580: Does this run before or after the real `/var` is available?

jprvita: Yes, `/var` is mounted by `ostree-prepare-root.service`, ordered `Before=initrd-root-fs.target`, whereas `eos-paygd-2.service` is ordered `After=initrd-root-fs.target`.

At this point in the code paygd has chroot'ed itself to `/sysroot` and was able to connect to D-Bus's system bus through a busy loop, so we now the real root is in a good state.

I think we can actually create the logdir earlier by pointing it to `/sysroot/var` when running from the initramfs, but I need to try it out.

eos-paygd/main.c
jprvita: Just personal preference of making it a number and shorter. I think `eos-paygd-20220418.log` reads better than `eos-paygd-2022-04-18.log` or `eos-paygd_2022-04-18.log` (the latter being slightly better, but I still prefer the current format.

eos-paygd/main.c
```diff
@@ -35,8 +35,95 @@
...
+  g_autofree gchar *tstamp = NULL;
+  g_autoptr(GError) error = NULL;
+
+  /* Build log file name with a date stamp so we get one file per day,
+   * ex., eos-paygd-20220418.log
+   */
+  now = g_date_time_new_now_local ();
+  tstamp = g_date_time_format (now, "%Y%m%d");
+  log_file_name = g_strconcat (LOGFILE_DIRNAME, "/", LOGFILE_BASENAME, "-",
+                               tstamp, "." LOGFILE_EXT, NULL);
```
jprvita, line 59:
```suggestion
                               tstamp, ".", LOGFILE_EXT, NULL);
```
Oops! (Still works due to the magical string literal concatenation performed by cpp).

jprvita:
Addressed a couple of the smaller problems highlighted in the comments above on the original commits, and added an extra commit that tries to start logging to this separate file as early as possible. This last part is still untested but ready for initial feedback.

jprvita:
> added an extra commit that tries to start logging to this separate file as early as possible. This last part is still untested but ready for initial feedback.

Ok, this did not work right away on my initial draft, and I don't think it is worth the debugging effort for the benefit of getting a few extra early debug messages. I'm reverting this extra commit, so the only changes from the last round were the fixes for comments in the PR (both mine and from Will).

jprvita:
> > added an extra commit that tries to start logging to this separate file as early as possible. This last part is still untested but ready for initial feedback.
> 
> Ok, this did not work right away on my initial draft, and I don't think it is worth the debugging effort for the benefit of getting a few extra early debug messages. I'm reverting this extra commit, so the only changes from the last round were the fixes for comments in the PR (both mine and from Will).

This is the comparison between the previously discussed revision and the current one: https://github.com/endlessm/eos-payg/compare/8e8c66583b374e0dc7da642cb44374d307b01408..T33220

jprvita:
> (I haven't been able to test the 90 days removal yet)

I've tested this by moving the clock fwd in the BIOS then waiting 15 minutes after boot for systemd-tmpfiles-clean.service to run, and it removed files older than 90 days, as expected.

eos-paygd/main.c
```diff
@@ -35,8 +35,92 @@
...
+   * ex., eos-paygd-20220418.log
+   */
+  now = g_date_time_new_now_local ();
+  tstamp = g_date_time_format (now, "%Y%m%d");
+  log_file_name = g_strconcat (LOGFILE_DIRNAME, "/", LOGFILE_BASENAME, "-",
+                               tstamp, ".", LOGFILE_EXT, NULL);
+  log_file = g_file_new_for_path (log_file_name);
+
+  *log_io = g_io_channel_new_file (log_file_name, "a", NULL);
+  if (!log_io)
```
wjt, line 62: You mean `if (!*log_io)` here. This function as written will either dereference a null pointer, or return TRUE unconditionally.

You could choose to instead return the `GIOChannel *` rather than `gboolean` and have no out parameter, which would make this bug harder to write.

jprvita: Oh, good catch, thanks! Yeah, I usually don't like out parameters because it is easy to overlook the dereferencing. There was something on the previous versions that made me think it was better that way, but with the way the code is right now I agree ditching the out parameter entirely.

eos-paygd/main.c
```diff
@@ -35,8 +35,90 @@
...
+  g_autoptr(GFile) log_file = NULL;
+  g_autoptr(GDateTime) now = NULL;
+  g_autofree gchar *tstamp = NULL;
+
+  /* Build log file name with a date stamp so we get one file per day,
+   * ex., eos-paygd-20220418.log
+   */
+  now = g_date_time_new_now_local ();
+  tstamp = g_date_time_format (now, "%Y%m%d");
+  log_file_name = g_strconcat (LOGFILE_DIRNAME, "/", LOGFILE_BASENAME, "-",
```
pwithnall, line 57: Technically this should be `g_build_filename()` rather than `g_strconcat()`, for portability to other platforms and for the case where the file system encoding is not UTF-8 (which is possible but vanishingly rare). Just an fyi, though, this isn’t worth changing now. The filename would have to be constructed separately (using `g_strconcat()`) before building the path.

eos-paygd/main.c
```diff
@@ -35,8 +35,90 @@
...
+                               tstamp, ".", LOGFILE_EXT, NULL);
+  log_file = g_file_new_for_path (log_file_name);
+
+  /* We can't log an error here if this fails, as it would cause infinite
+   * recursion */
+  return g_io_channel_new_file (log_file_name, "a", NULL);
+}
+
+/* we can't use g_log_writer_default_would_drop until glib 2.68, which will be
```
pwithnall, line 66: Nitpick: I would have used (pseudocode) `#if GLIB_CHECK_VERSION(2, 68, 0) { return g_log_writer_default_would_drop (…) } #else { … simplified fallback … }` so that it’s easier to find and drop the fallback code in future (by grepping for `GLIB_CHECK_VERSION`). Again, just an fyi, not worth changing now.

eos-paygd/main.c
```diff
@@ -35,8 +35,90 @@
...
+{
+  GIOChannel *log_io = NULL;
+
+  g_return_val_if_fail (fields != NULL, G_LOG_WRITER_UNHANDLED);
+  g_return_val_if_fail (n_fields > 0, G_LOG_WRITER_UNHANDLED);
+
+  if (should_drop_message (log_level))
+    return G_LOG_WRITER_HANDLED;
+
+  log_io = open_log_file ();
```
pwithnall, line 108: Should this be opening the log file, writing to it, and closing it again every time a message is logged? That seems costly.

Would it be better to keep a persistent handle to the open log file?

eos-paygd/main.c
```diff
@@ -487,6 +569,23 @@ main (int   argc,
       eospayg_efi_root_pivot ();
     }
 
+  /* Use /bin/mkdir instead of mkdir() to ensure the mode is unaffected by
+   * the process's umask.
+   */
+  system_ret = system ("/bin/mkdir -p --mode=755 " LOGFILE_DIRNAME);
```
pwithnall, line 575: Is this not done by the systemd-tmpfiles snippet? Sorry if that’s already been covered in a discussion; I only skimmed through them.

*Edit*: Oh, I see it was [discussed here](https://github.com/endlessm/eos-payg/pull/104#discussion_r854633924). Perhaps the details about startup ordering and dependencies on `RequiresMountsFor=` are worth a code comment here?

eos-paygd/main.c
pwithnall: Actually, the return value from `g_strconcat()` is leaked here, so this does need some tweaking.

pwithnall (commented):
Here are a few post-hoc review comments. Sorry for nitpicking.

I’ve tried to address them in https://github.com/endlessm/eos-payg/pull/105. I have not addressed the comment about reopening the log file for each message being logged, as that seemed like it needed a bit more discussion before being changed (if at all).

eos-paygd/main.c
jprvita: Re-opening the file every time is the way we designed cheap log rotation: we can't use `logrotate` because eos-payg can't receive signals in "phase-4 mode", and while there are other design alternatives (use logrotate and monitor the fd for renames with inotify, monitor the date or file size and close + re-open at that point), they are all more complicated and this seems good enough for what we need while remaining relatively pretty simple.

This designed choice is mentioned in the commit message of https://github.com/endlessm/eos-payg/commit/4a98f292cce82d260423aef614f7aede580a58e1, but it seems like we should also add it as a comment in the code.

pwithnall: :+1:

PR #105, opened 2022-04-25
Various small logging cleanups

pwithnall:
Based on review comments in https://github.com/endlessm/eos-payg/pull/104.

I have not tested this at all (even to compile it).


jprvita:
Oh, and I've added an extra commit adding a comment in the code on why we re-open the log file for every message (it was only in the commit message before).

pwithnall:
> Oh, and I've added an extra commit adding a comment in the code on why we re-open the log file for every message (it was only in the commit message before).

:ok_hand: :ok_hand: :ok_hand:

PR #106, opened 2022-11-17
jenkins: Use public eos container image

dbnicholson:
https://phabricator.endlessm.com/T34108


wjt:
retest this please

Part of the build process clones a Git submodule from GNOME GitLab, which was broken. It is now fixed, but this is an interesting case of our builds being contingent on an external service!

PR #107, opened 2022-11-18
service: Pass current time to provider on clock jumps

jprvita:
When a clock jump is detected, we already need to get the current wallclock time to calculate the jump size. Let's pass this information to the provider, so it can better decide how to handle the jump.

Also update the "manager" provider to use the time passed by caller in its wallclock_time_changed implementation instead of calculating again.

https://phabricator.endlessm.com/T33873


jprvita:
This is part of https://github.com/endlessm/eos-payg-nonfree/pull/107 -- it is a requirement for that PR to build, and we should only merge this one if we also approve and plan to merge that one.

PR #108, opened 2023-03-07
provider: Document now_sec on epg_provider_shutdown_finish

jprvita:
This parameter was added a few commits ago, but the documentation was missed.

### DIFF
--- a/libeos-payg/provider.c
+++ b/libeos-payg/provider.c
@@ -375,6 +375,8 @@ epg_provider_shutdown_finish (EpgProvider   *self,
  * @self: an #EpgProvider
  * @delta: the amount by which the clock changed in seconds, which can be
  * positive or negative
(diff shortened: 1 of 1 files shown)


libeos-payg/provider.c
```diff
@@ -375,6 +375,7 @@ epg_provider_shutdown_finish (EpgProvider   *self,
  * @self: an #EpgProvider
  * @delta: the amount by which the clock changed in seconds, which can be
  * positive or negative
+ * @now_secs: the time that the clock jump occurred, in unix seconds
```
pwithnall, line 378: Nitpick: Is that time given in the frame of reference of the old clock, or the new one?

Otherwise LGTM, and you could probably go ahead and merge it once you’ve addressed/responded to this comment, without further review

jprvita: Oh, good catch! Thanks

pwithnall (approved):

PR #109, opened 2023-04-12
manager: Fix credit consumption on clock jumps

jprvita:
https://phabricator.endlessm.com/T34550

### DIFF
--- a/libeos-payg/manager.c
+++ b/libeos-payg/manager.c
@@ -718,8 +718,7 @@ check_expired_cb (gpointer user_data)
   return G_SOURCE_CONTINUE;
 }
 
-/* Set the #EpgManager:expiry-time to
- * `MIN (G_MAXUINT64, MAX (@now, #EpgManager:expiry-time) + @span)` and set
+/* Set the #EpgManager:expiry-time to `MIN (G_MAXUINT64, @now + @span)` and set
  * the #GSource expiry timer to the new expiry time. Everything is handled in
  * seconds.
  *
@@ -739,12 +738,9 @@ set_expiry_time (EpgManager *self,
    * adjustments have no effect */
   guint64 old_expiry_time_secs = self->expiry_time_secs;
 
-  /* If the old PAYG code had expired, start from @now_secs; otherwise start
-   * from the current expiry time (in the future). */
-  guint64 base_secs = (now_secs < old_expiry_time_secs) ? old_expiry_time_secs : now_secs;
-
   /* Clamp to the end of time instead of overflowing. */
-  self->expiry_time_secs = (base_secs <= G_MAXUINT64 - span_secs) ? base_secs + span_secs : G_MAXUINT64;
+  if (!g_uint64_checked_add (&self->expiry_time_secs, now_secs, span_secs))
+    self->expiry_time_secs = G_MAXUINT64;
 
   /* Set the expiry timer. epg_clock_source_new_seconds() takes a #guint, and
    * @span_secs is a #guint64 so clamp to G_MAXUINT */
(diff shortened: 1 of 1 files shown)


libeos-payg/manager.c
```diff
@@ -739,12 +738,8 @@ set_expiry_time (EpgManager *self,
    * adjustments have no effect */
   guint64 old_expiry_time_secs = self->expiry_time_secs;
 
-  /* If the old PAYG code had expired, start from @now_secs; otherwise start
-   * from the current expiry time (in the future). */
-  guint64 base_secs = (now_secs < old_expiry_time_secs) ? old_expiry_time_secs : now_secs;
-
   /* Clamp to the end of time instead of overflowing. */
-  self->expiry_time_secs = (base_secs <= G_MAXUINT64 - span_secs) ? base_secs + span_secs : G_MAXUINT64;
+  self->expiry_time_secs = (now_secs <= G_MAXUINT64 - span_secs) ? now_secs + span_secs : G_MAXUINT64;
```
dbnicholson, line 742: Not that you necessarily want to change the code now, but you can use [`g_uint64_checked_add`](https://docs.gtk.org/glib/func.uint64_checked_add.html) for these types of things:

```
if (!g_uint64_checked_add (&self->expiry_time_secs, now_secs, span_secs))
  self->expiry_time_secs = G_MAXUINT64;
```

wjt: I think this is worth changing just for readability

libeos-payg/manager.c
```diff
@@ -718,8 +718,7 @@ check_expired_cb (gpointer user_data)
   return G_SOURCE_CONTINUE;
 }
 
-/* Set the #EpgManager:expiry-time to
- * `MIN (G_MAXUINT64, MAX (@now, #EpgManager:expiry-time) + @span)` and set
+/* Set the #EpgManager:expiry-time to `MIN (G_MAXUINT64, @now + @span)` and set
```
wjt, line 721: `@now` and `@span` parameters have since been renamed, these should change too.

libeos-payg/manager.c
```diff
@@ -1702,12 +1702,8 @@ epg_manager_wallclock_time_changed (EpgProvider *provider,
...
        * terms of CLOCK_BOOTTIME, so we get the time here again. */
       guint64 now_secs = epg_clock_get_time (self->clock);
-
-      /* FIXME: This actually extends the expiration rather than reducing it
-       * but the Endless backend is basically deprecated anyway in favor of
-       * Angaza.
-       */
-      set_expiry_time (self, NULL, FALSE, now_secs, (guint64)delta);
+      set_expiry_time (self, NULL, FALSE, now_secs,
+                       self->expiry_time_secs - now_secs - (guint64)delta);
```
wjt, line 1706: If `self->expiry_time_secs - now_secs` is negative, this will underflow.

Ignoring that, if `(self->expiry_time_secs - now_secs)` is less than `delta` (i.e. the delta consumes all remaining time) then it will also underflow.

libeos-payg/manager.c
jprvita: Oh, it took me a minute to realize that yes, it has changed some time in the past, not as part of this PR :smile:

I'm adding an extra commit updating the comment.

libeos-payg/manager.c
jprvita: Personally, I would have to look that function up while reading the code, but I agree that the suggestion is more readable once you know it. I'm adding a fixup commit.

libeos-payg/manager.c
jprvita: Both cases should now be handled in the new commits.

wjt (approved):

PR #110, opened 2023-04-14
Support 31-day codes

wjt:
Paying on the same calendar day each month is predictable and standard
for other subscription models. But previously, doing this would either
mean issuing two codes in months with 31 days (one for 30 days, another
for 1 day) or the user running out of credit for a day.

By supporting 31-day codes, it is possible to either issue a 30-day or
31-day code depending on the month, or (more likely) to simply issue a
31-day code every month, leaving the user with 7-8 days' extra credit at
the end of each year.

(We could in principle add 28-day and 29-day codes to support February
and lunar calendars "natively", but there are only 5 more bits available
in this bitfield.)

https://phabricator.endlessm.com/T33556

### DIFF
--- a/eos-payg-generate/main.c
+++ b/eos-payg-generate/main.c
@@ -67,6 +67,7 @@ periods[] =
     { EPC_PERIOD_13_DAYS, "13d", N_("13 days") },
     { EPC_PERIOD_14_DAYS, "14d", N_("14 days") },
     { EPC_PERIOD_30_DAYS, "30d", N_("30 days") },
+    { EPC_PERIOD_31_DAYS, "31d", N_("31 days") },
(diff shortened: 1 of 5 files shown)


wjt:
I found this code lying around uncommitted in my working copy when I went to check out #109!

jprvita (approved):
LGTM. Has this been tested, other than the units test, to make sure it does not break other parts of the stack (ex., the UI messages)? I don't expect it causing problems on those other bits, but I don't want new fires to put out :)

wjt:
OK, I have now done one more thing:

```
./eos-payg-generate/eos-payg-generate-1 /etc/os-release 31d
```

This successfully generated a bunch of codes.

Personally I think that, while we are landing some changes to this component anyway, now would be a good time to ship this change. I think it is safe!

PR #111, opened 2023-04-27
manager: Adjust credit on systems with way too much

jprvita:
Due to a software bug, some systems ended up with aproximately 50 years of credit. To recover from this situation, this commit adjusts systems with more than 5 years of credit to have 31 days of credit. These values were chosen because we do not expect a system to have 5+ years of credit by valid means, and because 31 days is the standard payment period with deployments using the manager (aka phase1) provider.

https://phabricator.endlessm.com/T34550

### DIFF
--- a/libeos-payg/manager.c
+++ b/libeos-payg/manager.c
@@ -1419,7 +1419,29 @@ file_load_cb (GObject      *source_object,
           if (unaccounted_time > self->last_save_expiry_secs)
             set_expiry_time (self, task, TRUE, now_secs, 0);
           else
-            set_expiry_time (self, task, TRUE, now_secs, self->last_save_expiry_secs - unaccounted_time);
+            {
+              guint64 YEAR_SECS = 365 * 24 * 60 * 60;
+              guint64 BOGUS_CREDIT_LOWER_LIMIT_SECS = 5 * YEAR_SECS;
+              guint64 BOGUS_CREDIT_UPPER_LIMIT_SECS = 100 * YEAR_SECS;
+              guint64 credit_secs = self->last_save_expiry_secs - unaccounted_time;
+
+              /* No system should have more than 5 years of credit, except if
+               * they were permanently unlocked, at which point G_MAXUINT64
+               * seconds (a bit over 584 billion years) were added to their
+               * credit.
+               * If a system in the field has over 5 years but less than 100
+               * years of credit it is certainly due to a bug (T34550). Reset
+               * the credit to 31 days, which is the typical payment period. */
+              if (BOGUS_CREDIT_LOWER_LIMIT_SECS < credit_secs &&
+                  credit_secs < BOGUS_CREDIT_UPPER_LIMIT_SECS)
+                {
+                  g_message ("Detected system with too much credit (%" G_GUINT64_FORMAT "), resetting to 31 days.",
+                             credit_secs);
+                  credit_secs = 31 * 24 * 60 * 60;
+                }
+
+              set_expiry_time (self, task, TRUE, now_secs, credit_secs);
+            }
         }
     }
 
--- a/libeos-payg/tests/manager.c
+++ b/libeos-payg/tests/manager.c
@@ -693,6 +693,76 @@ test_manager_add_infinite_code (Fixture      *fixture,
   g_assert_cmpuint (G_MAXUINT64, ==, expiry_after_reload);
 }
 
+/* test_manager_over_5_years_of_credit:
+ *
+ * Tests that if the expiry time of a system is over 5 years in the future, but
+ * less than 100 years (which would mean permanently unlocked), the expiry time
+ * is reset to 31 days in the future after a reboot (see ticket T34550 for
+ * rationale).
+ */
+static void
+test_manager_over_5_years_of_credit (Fixture      *fixture,
+                                     gconstpointer data)
+{
+  manager_new (fixture);
+  EpcCode code;
+  g_autofree gchar *code_str = NULL;
+  g_autoptr(GError) error = NULL;
+  gint64 time_added = 0;
+  guint64 expiry_before_code, expiry_after_code, expiry_after_reload;
+  gboolean ret;
+
+  guint64 ONE_YEAR_SECS = 365 * 24 * 60 * 60;
+
+  expiry_before_code = epg_provider_get_expiry_time (fixture->provider);
+  g_assert_cmpuint (G_MAXUINT64, !=, expiry_before_code);
+
+  /* Apply 5 one-year codes */
+  for (guint i = 0; i < 5; i++)
+    {
+      code = epc_calculate_code (EPC_PERIOD_365_DAYS, fixture->next_counter++, fixture->key, &error);
+      g_assert_no_error (error);
+
+      code_str = epc_format_code (code);
+      ret = epg_provider_add_code (fixture->provider, code_str, &time_added, &error);
+      g_assert_no_error (error);
+      g_assert_true (ret);
+
+      g_assert_cmpint (ONE_YEAR_SECS, ==, time_added);
+      g_clear_pointer (&code_str, g_free);
+
+      /* Tear down and restart after every code, to avoid a race condition when
+       * writing to disk */
+      ret = shutdown (fixture, &error);
+      g_assert_no_error (error);
+      g_assert_true (ret);
+      manager_new (fixture);
+    }
+
+  expiry_after_code = epg_provider_get_expiry_time (fixture->provider);
+  g_assert_cmpint (expiry_before_code + 5 * ONE_YEAR_SECS, ==, expiry_after_code);
+
+  /* Apply a 5-second code get past the 5-year threshold */
+  g_clear_pointer (&code_str, g_free);
+  code_str = get_next_code (fixture);
+  ret = epg_provider_add_code (fixture->provider, code_str, &time_added, &error);
+  g_assert_no_error (error);
+  g_assert_true (ret);
+
+  expiry_after_code = epg_provider_get_expiry_time (fixture->provider);
+  g_assert_cmpint (5, ==, time_added);
+  g_assert_cmpint (expiry_before_code + 5 * ONE_YEAR_SECS + 5, ==, expiry_after_code);
+
+  /* Tear down and restart */
+  ret = shutdown (fixture, &error);
+  g_assert_no_error (error);
+  g_assert_true (ret);
+
+  manager_new (fixture);
+  expiry_after_reload = epg_provider_get_expiry_time (fixture->provider);
+  g_assert_cmpuint (expiry_before_code + 31 * 24 * 60 * 60, ==, expiry_after_reload);
+}
+
 /* test_manager_error_malformed:
  *
  * Tests that entering a totally malformed code does not extend the expiry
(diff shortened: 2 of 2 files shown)


libeos-payg/tests/manager.c
```diff
@@ -637,21 +637,25 @@ test_manager_add_save_reload (Fixture *fixture,
 /* test_manager_add_infinite_code:
  *
  * Tests that applying an infinite code sets the expiry time to infinitely far
- * in the future, and that this survives both a restart and adding a
- * non-infinite code afterwards.
+ * in the future, and that this survives adding a non-infinite code afterwards.
+ *
+ * After a reboot, the expiry time should be reset to 31 days in the future
```
wjt, line 642: Hmm. Don't we use an infinite code as the final unlock code when the loan is paid off?

Can we adjust the logic to "more than 5 years but less than maxint"?

jprvita: Oh, I think you are right -- good catch.

jprvita: oh, I think that will actually not work, because the time will continue to pass, so it will only be maxint at the second the final unlock code is added. Perhaps what we should do here is keep these changes as-is, but also implement disabling PAYG when a final unlock code is added (which can continue to be maxint, but we just treat it differently).

wjt: Ugh, well… there are already systems in the field which are fully paid off, I think… They would be very sad to lose all their credit.

jprvita: Ok, so maybe the heuristics here is to reset if credit is over 5 years but under 60 (or 100) years? maxint seconds should be a bit over 584 billion years.

wjt: That should indeed leave a wide margin of error! Good idea.

jprvita: Ok, I think I have that written correctly, but I'm having trouble with a new test I have added to check the `credit > 100 years` situation, in https://github.com/endlessm/eos-payg/commit/c2933a121b333c08aa8ae8fec4a62a8ff46bf4bd. It passes some times and fails some times, and every time after "reboot" it has a different amount of credit which sometimes fall inside the window we are targeting and other times not . I'll get back to this tomorrow, but any ideas on where this non-determinism could be coming from would be appreciated.

jprvita:
bot, test this please

libeos-payg/tests/manager.c
wjt: No, I can't see what's wrong with that I'm afraid :(

wjt: I guess what I would probably do next is add some stuff to the test suite to test what has been persisted to disk between shutdown and startup, and maybe get some more logging into the startup code that reads those files and eventually calculates the credit remaining.

wjt: > Oh, I think that will actually not work, because the time will continue to pass, so the amount of credit (or seconds left) will only be maxint at the second the final unlock code is added.

Please humour me here – is this actually true? Can we test it? I have a dim memory that the expiry time was stored in terms of credit remaining, and that it would stay at maxuint64 once set to that.

Failing that – I checked our gnome-shell fork and it considers any time above MAXUINT32 (not 64) to be "infinite". Does raising the threshold to that work?

Finally, a bad idea I had while out walking: how many computers are affected by this bug? Could we instead just arrange to run `eos-payg-ctl clear-code` on each of theM?

wjt: Or a crude option:

Or, an even worse idea:

- If a stamp file does not exist
- And the image product & personality matches what we expect – we know no machines in that cohort are fully paid off
- And the credit is over 31 days
- then
   - Set it to 31 days
   - touch the stamp file

jprvita: > > Oh, I think that will actually not work, because the time will continue to pass, so the amount of credit (or seconds left) will only be maxint at the second the final unlock code is added.
>
> Please humour me here – is this actually true? Can we test it? I have a dim memory that the expiry time was stored in terms of credit remaining, and that it would stay at maxuint64 once set to that.

What is stored to disk is a timestamp of the save in terms of wallclock time and the number of seconds (so, credit) remaining. But on loading there is no special treatment of maxuint64, and even if there was, for this to work we would have to somehow stop to consume credits after a maxuint64 code was entered, because after entering the code the time continues to pass and on the next save (on shutdown or the next clock jump) the new saved expiry seconds will be a little less than maxuint64.

I think here is the reason to store credits separate from time, if we ever have to write a new PAYG provider (potentially for open nexus). That is also how the Angaza library is designed.

> Failing that – I checked our gnome-shell fork and it considers any time above MAXUINT32 (not 64) to be "infinite". Does raising the threshold to that work?

That makes sense, but I believe it would affect the test in the same way, and to test MAXUINT32 < credit < MAXUINT64 I would have to add 140 * 1-year codes (no too different than adding 100 of those).

> Finally, a bad idea I had while out walking: how many computers are affected by this bug? Could we instead just arrange to run `eos-payg-ctl clear-code` on each of them?

Maybe, but that is a question for Ale. 

> Or a crude option:
> 
>     * If a stamp file does not exist
> 
>     * And the image product & personality matches what we expect – we know no machines in that cohort are fully paid off
> 
>     * And the credit is over 31 days
> 
>     * then
>       
>       * Set it to 31 days
>       * touch the stamp file

I think that is a bit too risky -- if the file gets deleted for some reason and the user enters a code early, we're removing their credit. I think what we currently have, `if (5_YEARS < credit < 100_YEARS_OR_MAXUNIT32) { reset_to_31d(); }`, is a better solution. And I believe the code is actually working, just the new test is failing for some reason.

> I guess what I would probably do next is add some stuff to the test suite to test what has been persisted to disk between shutdown and startup, and maybe get some more logging into the startup code that reads those files and eventually calculates the credit remaining.

I think that is where our answer lies.

libeos-payg/manager.c
```diff
@@ -1416,7 +1416,29 @@ file_load_cb (GObject      *source_object,
...
+               * they were permanently unlocked, at which point G_MAXUINT64
+               * seconds (a bit over 584 billion years) were added to their
+               * credit.
+               * If a system in the field has over 5 years but less than 100
+               * years of credit it is certainly due to a bug (T34550). Reset
+               * the credit to 31 days, which is the typical payment period. */
+              if (BOGUS_CREDIT_LOWER_LIMIT_SECS < credit_secs &&
+                  credit_secs < BOGUS_CREDIT_UPPER_LIMIT_SECS)
+                {
+                  g_message ("Detected system with too much credit (%" G_GUINT64_FORMAT "), resetting to 31 days.",
```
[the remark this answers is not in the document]
jprvita: I wouldn't add the `seconds`, the message is already pretty long and I'm not sure it adds much value.

libeos-payg/tests/manager.c
wjt: I looked into this a bit and recalled that, whenever a code is added, an async operation is kicked off which, in three background threads pulled from the thread pool, writes three files. If you add 11 (or 5 – the 5-year test failed in the same way, just less often) codes, all 11 × 3 = 33 operations race one another. There's no guarantee that the most recent ones win.

On shutdown, if there is **no** save in flight, a single fresh save is kicked off and waited for. But if at least one is in flight, all of them completing is deemed sufficient.

So I tried the following:

```patch
diff --git a/libeos-payg/tests/manager.c b/libeos-payg/tests/manager.c
index 7fcea36..305209c 100644
--- a/libeos-payg/tests/manager.c
+++ b/libeos-payg/tests/manager.c
@@ -728,6 +728,13 @@ test_manager_over_5_years_of_credit (Fixture      *fixture,
       g_assert_true (ret);
       g_assert_cmpint (ONE_YEAR_SECS, ==, time_added);
       g_clear_pointer (&code_str, g_free);
+
+        /* Tear down and restart. */
+  ret = shutdown (fixture, &error);
+  g_assert_no_error (error);
+  g_assert_true (ret);
+
+  manager_new (fixture);
     }
 
   expiry_after_code = epg_provider_get_expiry_time (fixture->provider);
@@ -788,6 +795,13 @@ test_manager_over_100_years_of_credit (Fixture      *fixture,
       g_assert_true (ret);
       g_assert_cmpint (ONE_YEAR_SECS, ==, time_added);
       g_clear_pointer (&code_str, g_free);
+
+        /* Tear down and restart. */
+  ret = shutdown (fixture, &error);
+  g_assert_no_error (error);
+  g_assert_true (ret);
+
+  manager_new (fixture);
     }
 
   expiry_after_code = epg_provider_get_expiry_time (fixture->provider);
```

which I expected to mask the problem by serialising the writes to disk. It completely fixes `over-5-years` but of course doesn't fix `over-100-years` because on the 6th and subsequent iterations the manager finds itself in the over-5-years-but-less-than-100-years case and sets the credit back to 31 days.

wjt: Then I tried a patch which adds an `EPC_PERIOD_110_YEARS` so that the test can avoid the loop and the race. The 100-years test passed 10,000 times in a loop. (By contrast it previously failed essentially every time, as you found.)

jprvita: > I looked into this a bit and recalled that, whenever a code is added, an async operation is kicked off which, in three background threads pulled from the thread pool, writes three files. If you add 11 (or 5 – the 5-year test failed in the same way, just less often) codes, all 11 × 3 = 33 operations race one another. There's no guarantee that the most recent ones win.
> 
> On shutdown, if there is **no** save in flight, a single fresh save is kicked off and waited for. But if at least one is in flight, all of them completing is deemed sufficient.

Oh! That explains why the assertion after adding all the codes does not fail. That is unfortunate, although I understand this does not matter in real-world use.

> 
> So I tried the following:
> 
> ```diff
> diff --git a/libeos-payg/tests/manager.c b/libeos-payg/tests/manager.c
> index 7fcea36..305209c 100644
> --- a/libeos-payg/tests/manager.c
> +++ b/libeos-payg/tests/manager.c
> @@ -728,6 +728,13 @@ test_manager_over_5_years_of_credit (Fixture      *fixture,
>        g_assert_true (ret);
>        g_assert_cmpint (ONE_YEAR_SECS, ==, time_added);
>        g_clear_pointer (&code_str, g_free);
> +
> +        /* Tear down and restart. */
> +  ret = shutdown (fixture, &error);
> +  g_assert_no_error (error);
> +  g_assert_true (ret);
> +
> +  manager_new (fixture);
>      }
>  
>    expiry_after_code = epg_provider_get_expiry_time (fixture->provider);
> @@ -788,6 +795,13 @@ test_manager_over_100_years_of_credit (Fixture      *fixture,
>        g_assert_true (ret);
>        g_assert_cmpint (ONE_YEAR_SECS, ==, time_added);
>        g_clear_pointer (&code_str, g_free);
> +
> +        /* Tear down and restart. */
> +  ret = shutdown (fixture, &error);
> +  g_assert_no_error (error);
> +  g_assert_true (ret);
> +
> +  manager_new (fixture);
>      }
>  
>    expiry_after_code = epg_provider_get_expiry_time (fixture->provider);
> ```
> 
> which I expected to mask the problem by serialising the writes to disk. It completely fixes `over-5-years` but of course doesn't fix `over-100-years` because on the 6th and subsequent iterations the manager finds itself in the over-5-years-but-less-than-100-years case and sets the credit back to 31 days.

Makes sense, I've fixed the 5-year test that way.

Then I spent some time trying to force a save on shutdown, but that bit of code is not the most straightforward and I'm afraid to introduce new bugs if we try to change it too much. So I abandoned that idea.

Then I got a bit creative and tried to simulate the 100+ years of credit by manipulating the clock instead of adding more codes. It works most of the time, but it is still failing some times. I'll investigate a bit more, but if that does not work I think we should either add an extra code just for testing, and maybe change from 100 years to maxuint32 for this special code, or simply drop the test (I'm not sure this test adds much value given we already have one that adds the infinite code)

wjt: I think we should add the test but marked as expected to fail (if GLib supports that) or commented out (if not) with a FIXME explaining why.

jprvita: > I think we should add the test but marked as expected to fail (if GLib supports that) or commented out (if not) with a FIXME explaining why.

Done! I did not have much luck with `g_test_skip`, so I just commented the whole test implementation out.

wjt (approved):

PR #114, opened 2024-01-18
Build-depend on systemd

wjt:
As described in the commit message, this is needed because two paths are looked up from `systemd.pc`. Evidently the `systemd` package is installed in the Jenkins build environment because 8f0ba1bf2427e4329369e60eefa54ce9773a35a2 builds fine on Jenkins; and evidently it is present in the OBS build environment on 5.1 because that same change built fine on `eos:5.1:endless`; but the build failed on `eos:master:endless`.

https://phabricator.endlessm.com/T35174

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -14,6 +14,7 @@ Build-Depends:
  libsystemd-dev,
  meson,
  python3-dbusmock,
+ systemd,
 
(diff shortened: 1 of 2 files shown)


starnight (approved):
Thanks!

PR #115, opened 2024-01-19
Require glib 2.74

wjt:
Spotted while working on https://github.com/endlessm/eos-meta/pull/783 and checking the details of the daemon's logging. Not strictly related but let's attach it to that ticket anyway:

https://phabricator.endlessm.com/T33896

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,11 +34,12 @@ again.
 Dependencies
 ============
 
- * gio-2.0 ≥ 2.54
- * glib-2.0 ≥ 2.54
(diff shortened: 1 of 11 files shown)


starnight (changes requested):
Thanks!

But, please also update the **Dependencies** in README.md.
After that, feel free to merge it!

wjt:
I updated the README to remove the specific versions (which were already wrong). Personally I'm of the opinion that listing that kind of stuff in the README is just guaranteed to become out of date.

PR #118, opened 2024-07-05
efi: Simplify closing fds

wjt:
[libglnx](https://gitlab.gnome.org/GNOME/libglnx) is a helper library designed to be embedded into another project as a git submodule. Among other things, it provides an attribute you can attach to a variable holding a file descriptor, which calls `close()` on it when it goes out of scope if it is non-negative. Any error from `close()` is ignored, just like our existing code, except if it is `EBADF`, which indicates a programming error.

Add a libglnx submodule, and use `glnx_autofd` throughout our EFI code. Most cases are simple but `eospayg_efi_init()` is a little bit more involved.

This will make it easier to propagate errors to the caller in a future commit.

Additionally, fix a couple of errors in docstrings.

https://phabricator.endlessm.com/T35459

### DIFF
--- a/libeos-payg/efi.c
+++ b/libeos-payg/efi.c
@@ -14,6 +14,7 @@
 #include <unistd.h>
 #include <sys/ioctl.h>
 #include <libeos-payg/efi.h>
+#include <libglnx.h>
 
 #define EOSPAYG_GUID         "d89c3871-ae0c-4fc5-a409-dc717aee61e7"
 #define GLOBAL_VARIABLE_GUID "8be4df61-93ca-11d2-aa0d-00e098032b8c"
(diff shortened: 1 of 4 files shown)


libeos-payg/efi.c
```diff
@@ -527,7 +527,7 @@ eospayg_efi_securebootoption_disabled (void)
  * space, but the kernel has a placeholder file for it due to a failed
  * write.
  *
- * Returns: size of the PK efi variable including EFI overhead, or -1 if missing
+ * Returns: size of the PK EFI variable, excluding the 4-byte attribute overhead, or -1 if missing
```
starnight, line 521: Do we have 80 characters rule here?  Others are LGTM!
If it is okay, feel free to merge it.

wjt: It's not documented but you're right: most of the rest of the file, and the rest of this docstring, has a column limit. I amended this commit to wrap this line. Thansk

PR #119, opened 2024-07-08
libeos-payg: Build libeos-payg as both shared & static libs

starnight:
To build eos-payg-nonfree against libeos-payg static lib for avoiding API/ABI issue, build and install libeos-payg as both shared & static libs.

https://phabricator.endlessm.com/T35437

### DIFF
--- a/debian/libeos-payg-1-dev.install
+++ b/debian/libeos-payg-1-dev.install
@@ -1,3 +1,4 @@
+usr/lib/*/eos-payg-1/libeos-payg-1.a
 usr/lib/*/pkgconfig/eos-payg-1.pc
 usr/include/eos-payg-1/libeos-payg
 
--- a/debian/rules
+++ b/debian/rules
@@ -8,7 +8,8 @@ override_dh_auto_configure:
 	dh_auto_configure \
 		-- \
 		-Dinstalled_tests=true \
-		-Ddefault_library=static \
+		-Ddefault_library=both \
+		-Dlibgsystemservice:default_library=static \
 		-Dtests=unsafe
 		$(NULL)
 
--- a/libeos-payg/meson.build
+++ b/libeos-payg/meson.build
@@ -51,7 +51,7 @@ libeos_payg_resources = gnome.compile_resources(
   c_name : 'epg',
 )
 
-libeos_payg = shared_library(eos_payg_api_name,
+libeos_payg = library(eos_payg_api_name,
   libeos_payg_sources + libeos_payg_headers + libeos_payg_resources,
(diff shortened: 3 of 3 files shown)


debian/eos-paygd.install
```diff
@@ -1,6 +1,7 @@
 lib/systemd/system/eos-paygd.service
 lib/systemd/system/eos-paygd-2.service
 usr/lib/tmpfiles.d/eos-paygd.conf
+usr/lib/*/eos-payg-1/libeos-payg-1.a
```
wjt, line 4: This should be part of the `libeos-payg-1-dev` package, so listed in `libeos-payg-1-dev.install`. It doesn't need to be installed on end-user systems, which `eos-paygd` is.

wjt (changes requested):

debian/eos-paygd.install
starnight: ya! sure!  fixed in the new commit.

wjt (commented):
On a hunch I tested building the package locally and it fails:

```
dpkg-shlibdeps: error: cannot find library libgsystemservice-0.so.0 needed by debian/eos-paygd/usr/lib/x86_64-linux-gnu/eos-payg-1/libeos-payg-1.so (ELF format: 'elf64-x86-64' abi: '0201003e00000000'; RPATH: '')
dpkg-shlibdeps: error: cannot find library libgsystemservice-0.so.0 needed by debian/eos-paygd/usr/lib/x86_64-linux-gnu/eos-paygd1 (ELF format: 'elf64-x86-64' abi: '0201003e00000000'; RPATH: '/usr/lib/x86_64-linux-gnu/eos-payg-1')
dpkg-shlibdeps: error: cannot continue due to the errors listed above
Note: libraries are not searched in other binary packages that do not have any shlibs or symbols file.
To help dpkg-shlibdeps find private libraries, you might need to use -l.
dh_shlibdeps: error: dpkg-shlibdeps -Tdebian/eos-paygd.substvars debian/eos-paygd/usr/lib/x86_64-linux-gnu/eos-payg-1/libeos-payg-1.so debian/eos-paygd/usr/lib/x86_64-linux-gnu/eos-paygd1 returned exit code 2
dh_shlibdeps: error: Aborting due to earlier error
make: *** [debian/rules:30: binary] Error 2
dpkg-buildpackage: error: fakeroot debian/rules binary subprocess returned exit status 2
```

wjt:
I have found the problem, I'll push a fixup.

starnight:
ah!  Thanks!

wjt (approved):

PR #120, opened 2024-07-09
efi: Introduce eospayg_efi_setupmode()

starnight:
Introduce eospayg_efi_setupmode() to check the system boots with SetupMode, or not. It check the EFI variable:
SetupMode-8be4df61-93ca-11d2-aa0d-00e098032b8c for normal cases.

https://phabricator.endlessm.com/T34898

### DIFF
--- a/libeos-payg/efi.c
+++ b/libeos-payg/efi.c
@@ -370,6 +370,7 @@ eospayg_efi_var_delete_fullname (const char  *name,
                                  GError     **error)
 {
   g_return_val_if_fail (efi != NULL, FALSE);
+  g_return_val_if_fail (name != NULL, FALSE);
   g_return_val_if_fail (error == NULL || *error == NULL, FALSE);
 
   /* Make sure we never delete a non EOSPAYG_
@@ -404,6 +405,7 @@ eospayg_efi_var_delete (const char  *name,
                         GError     **error)
 {
   g_return_val_if_fail (efi != NULL, FALSE);
+  g_return_val_if_fail (name != NULL, FALSE);
   g_return_val_if_fail (error == NULL || *error == NULL, FALSE);
 
   g_autofree char *tname = eospayg_efi_name (name);
@@ -445,6 +447,7 @@ gboolean
 eospayg_efi_var_exists (const char *name)
 {
   g_return_val_if_fail (efi != NULL, FALSE);
+  g_return_val_if_fail (name != NULL, FALSE);
 
   return efi->exists (name);
 }
@@ -689,6 +692,7 @@ eospayg_efi_var_read_fullname (const char  *name,
                                GError     **error)
 {
   g_return_val_if_fail (efi != NULL, FALSE);
+  g_return_val_if_fail (name != NULL, FALSE);
   g_return_val_if_fail (expected_size >= -1, FALSE);
   g_return_val_if_fail (size != NULL, FALSE);
   g_return_val_if_fail (error == NULL || *error == NULL, FALSE);
@@ -727,6 +731,8 @@ eospayg_efi_var_read_fullname_boolean (const char  *name,
                                        GError     **error)
 {
   g_return_val_if_fail (efi != NULL, FALSE);
+  g_return_val_if_fail (name != NULL, FALSE);
+  g_return_val_if_fail (error == NULL || *error == NULL, FALSE);
 
   g_autofree unsigned char *content = NULL;
   int size;
@@ -764,6 +770,10 @@ eospayg_efi_var_read (const char  *name,
                       GError     **error)
 {
   g_return_val_if_fail (efi != NULL, FALSE);
+  g_return_val_if_fail (name != NULL, FALSE);
+  g_return_val_if_fail (expected_size >= -1, FALSE);
+  g_return_val_if_fail (size != NULL, FALSE);
+  g_return_val_if_fail (error == NULL || *error == NULL, FALSE);
 
   g_autofree char *tname = eospayg_efi_name (name);
 
--- a/libeos-payg/efi.c
+++ b/libeos-payg/efi.c
@@ -370,6 +370,7 @@ gboolean
 eospayg_efi_var_delete_fullname (const char  *name,
                                  GError     **error)
 {
+  g_return_val_if_fail (efi != NULL, FALSE);
   g_return_val_if_fail (error == NULL || *error == NULL, FALSE);
 
   /* Make sure we never delete a non EOSPAYG_
@@ -403,6 +404,7 @@ gboolean
 eospayg_efi_var_delete (const char  *name,
                         GError     **error)
 {
+  g_return_val_if_fail (efi != NULL, FALSE);
   g_return_val_if_fail (error == NULL || *error == NULL, FALSE);
 
   g_autofree char *tname = eospayg_efi_name (name);
@@ -443,6 +445,8 @@ efivarfs_exists (const char *name)
 gboolean
 eospayg_efi_var_exists (const char *name)
 {
+  g_return_val_if_fail (efi != NULL, FALSE);
+
   return efi->exists (name);
 }
 
@@ -549,6 +553,8 @@ eospayg_efi_secureboot_active (void)
   if (test_mode)
     return TRUE;
 
+  g_return_val_if_fail (efi != NULL, FALSE);
+
   g_autofree char *name = full_efi_name (GLOBAL_VARIABLE_GUID, "SecureBoot");
 
   return eospayg_efi_var_read_fullname_boolean (name, NULL);
@@ -563,6 +569,8 @@ eospayg_efi_secureboot_active (void)
 gboolean
 eospayg_efi_setupmode_active (void)
 {
+  g_return_val_if_fail (efi != NULL, FALSE);
+
   g_autofree char *name = full_efi_name (GLOBAL_VARIABLE_GUID, "SetupMode");
 
   return eospayg_efi_var_read_fullname_boolean (name, NULL);
@@ -582,6 +590,7 @@ eospayg_efi_setupmode_active (void)
 enum efivar_states
 eospayg_efi_secureboot_setup_active (void)
 {
+  g_return_val_if_fail (efi != NULL, FALSE);
   g_autofree char *name = full_efi_name (SECUREBOOT_SETUP_GUID,
                                          "SecureBootSetup");
   g_autoptr(GError) error = NULL;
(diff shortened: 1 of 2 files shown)


libeos-payg/efi.c
```diff
@@ -474,6 +474,26 @@ eospayg_efi_secureboot_active (void)
...
+ * Returns: %TRUE if booted with SetupMode, %FALSE otherwise
+ */
+gboolean
+eospayg_efi_setupmode_active (void)
+{
+  g_autofree char *tname = full_efi_name (GLOBAL_VARIABLE_GUID, "SetupMode");
+  g_autofree unsigned char *content = NULL;
+  int size;
+
+  content = efivarfs_read (tname, &size);
```
wjt, line 490: If you look at the other accessors for specific variables, such as `eospayg_efi_secureboot_active`, they have a branch for `test_mode` rather than unconditionally reading from the real efivarfs.

Personally I think this is wrong – they should go through the same vfunc as `eospayg_efi_var_read` in my opinion, not least so that code that calls them can be meaningfully tested – but this function should at least be consistent with the others and return a sensible value in tests.

starnight: I think `eospayg_efi_var_read` is unsuitable here.  Because, it invokes [eospayg_efi_name()](https://github.com/endlessm/eos-payg/blob/b3267a888bd00cce4da11026ecb58949fe73f666/libeos-payg/efi.c#L652C28-L652C44), which prepends the EFI variable name with `EOSPAYG_`.  Then, it finds the wrong one and cannot find the file.

wjt: You can't use `eospayg_efi_var_read` directly, that's true.

If you call `efivarfs_read` which directly reads from `efivarfs` then you need to check the `test_mode` variable and return some sensible value if it is true.

Or you can use the `efi->read` virtual function, like `eospayg_efi_var_read` does, which either reads from some fake test storage or from the real EFI variables depending on the flags passed to `eospayg_efi_init()` when that function was called.

wjt: > Because, it invokes [eospayg_efi_name()](https://github.com/endlessm/eos-payg/blob/b3267a888bd00cce4da11026ecb58949fe73f666/libeos-payg/efi.c#L652C28-L652C44), which prepends the EFI variable name with `EOSPAYG_`. Then, it finds the wrong one and cannot find the file.

Notice that there are two delete functions, eospayg_efi_var_delete and eospayg_efi_var_delete_fullname. You could add a `eospayg_efi_var_read_fullname`.

libeos-payg/efi.c
```diff
@@ -559,6 +559,26 @@ eospayg_efi_secureboot_active (void)
...
+ * Returns: %TRUE if booted with SetupMode, %FALSE otherwise
+ */
+gboolean
+eospayg_efi_setupmode_active (void)
+{
+  g_autofree char *tname = full_efi_name (GLOBAL_VARIABLE_GUID, "SetupMode");
+  g_autofree unsigned char *content = NULL;
+  int size;
+
+  content = efivarfs_read (tname, &size, NULL);
```
wjt, line 576: As discussed you need to either handle `test_mode` specially, or call `efi->read` which itself handles `test_mode` specially (assuming FALSE is the correct value for the test suite by default), or equivalently add `eospayg_efi_var_read_fullname` and call that.

libeos-payg/efi.c
```diff
@@ -559,6 +559,26 @@ eospayg_efi_secureboot_active (void)
...
+  if (!content || size != 1)
+    return FALSE;
+
+  return !!content[0];
```
wjt, line 580: This function's body is the same as eospayg_efi_secureboot_active just with a different variable name. Perhaps you could define a generic `eospayg_efi_read_boolean()` function instead.

libeos-payg/efi.c
```diff
@@ -579,6 +580,34 @@ eospayg_efi_setupmode_active (void)
   return !!content[0];
 }
 
+/* eospayg_efi_secureboot_setup_active:
+ *
+ * Check if the system was booted with SecureBoot Setup config
+ *
+ * Some system does not have EFI variable SecureBootSetup. So, this will return
+ * %FALSE, too. The caller has to check the @has, if it does not have EFI
+ * variable SecureBootSetup.
```
wjt, line 589: Personally I don't particularly like this API. I would prefer it to return a value from a 3-state enum. But if you prefer this API, please rename the out parameter to `found`.

libeos-payg/efi.c
```diff
@@ -579,6 +580,34 @@ eospayg_efi_setupmode_active (void)
...
+gboolean
+eospayg_efi_secureboot_setup_active (gboolean *has)
+{
+  g_autofree char *tname = full_efi_name (SECUREBOOT_SETUP_GUID,
+                                          "SecureBootSetup");
+  g_autofree unsigned char *content = NULL;
+  int size;
+
+  content = efivarfs_read (tname, &size, NULL);
+  *has = !errno;
```
wjt, line 602:
```c
g_autoptr(GError) error = NULL;
content = efivarfs_read (tname, &size, &error);
*has = g_error_matches (error, G_IO_ERROR, G_IO_ERROR_NOT_FOUND);
```

libeos-payg/efi.c
```diff
@@ -559,6 +560,54 @@ eospayg_efi_secureboot_active (void)
...
+ */
+gboolean
+eospayg_efi_secureboot_setup_active (gboolean *has)
+{
+  g_autofree char *tname = full_efi_name (SECUREBOOT_SETUP_GUID,
+                                          "SecureBootSetup");
+  g_autofree unsigned char *content = NULL;
+  int size;
+
+  content = efivarfs_read (tname, &size, NULL);
```
wjt, line 601: Again this needs to handle test mode, one way or another.

wjt (changes requested):
Needs to handle test mode; and some other improvements.

libeos-payg/efi.c
starnight: This is more readable.  Thanks

libeos-payg/efi.c
```diff
@@ -537,26 +544,60 @@ efivarfs_read (const char  *name,
...
 
-  /* In test mode let's pretend secure boot is enabled */
-  if (test_mode)
-    return TRUE;
+  g_autofree char *name = full_efi_name (GLOBAL_VARIABLE_GUID, "SecureBoot");
 
-  content = efivarfs_read (tname, &size, NULL);
-  if (!content || size != 1)
-    return FALSE;
+  return eospayg_efi_var_read_fullname_boolean (name, NULL);
```
wjt, line 558: This changes the behaviour of this function in tests. Previously in the test suite which calls `eospayg_efi_init (EOSPAYG_EFI_TEST_MODE, ...)` this function would unconditionally return TRUE. Now it will return FALSE.

libeos-payg/efi.c
```diff
@@ -537,26 +544,60 @@ efivarfs_read (const char  *name,
...
+ * Check if the system was booted with SecureBoot Setup config.
+ *
+ * However, some systems such as OLPC LEAP W502 do no have the SecureBootSetup
+ * EFI variable.
+ *
+ * Returns: one of the enum. %EFIVAR_NOT_EXIST if SecureBootSetup variable does
+ *          not exist, %EFIVAR_TRUE if system boots with set SecureBootSetup,
+ *          %EFIVAR_FALSE otherwise
+ */
+enum efivar_states
```
wjt, line 588: I know it was I who suggested the enum, but now that I see how it is used (`eospayg_efi_secureboot_setup_active () == EFIVAR_FALSE` at the only callsite) I wonder if it should in fact return a single boolean in a similar "oddly inverted" way to the function below, `eospayg_efi_securebootoption_disabled`.

libeos-payg/efi.c
```diff
@@ -669,6 +711,71 @@ eospayg_efi_var_read (const char  *name,
...
+{
+  g_return_val_if_fail (efi != NULL, FALSE);
+  g_return_val_if_fail (name != NULL, FALSE);
+  g_return_val_if_fail (error == NULL || *error == NULL, FALSE);
+
+  g_autofree unsigned char *content = NULL;
+  int size;
+
+  content = eospayg_efi_var_read_fullname (name, -1, &size, error);
+  if (!content || size < 1)
```
wjt, line 737: On this code path if `size` is less than 1 you are returning `FALSE` but not setting `error`. I guess it's also the case that `error` is not set when `content[0]` is zero...

Typically the convention is that a false-y return value should always set the error. In practice it doesn't matter because the only caller is inside this source file and it handles that case directly. Perhaps make this function `static`.

wjt: Actually if you check the implementation of `efivarfs_read` it is impossible for `size` to be zero but `content` to be non-NULL. Theoretically `test_read` can return zero `size` but non-NULL `content` but only if someone incorrectly writes an empty variable. According to other comments in this code, experimentation, the corresponding Windows API, and the [SetVariable](https://uefi.org/specs/UEFI/2.10/08_Services_Runtime_Services.html#setvariable) UEFI function, the variable data cannot be zero bytes long.

libeos-payg/efi.c
wjt: Either way is fine – it's just an idea.

starnight: hmm ...  For returning a single boolean, I think it will be unclear, if there is no SecureBootSetup EFI variable.

libeos-payg/efi.c
starnight: fixed in the new commit

libeos-payg/efi.c
wjt: I don't necessarily disagree, I'm just saying that there is another function in this same file that collapses the three states (true, false, not found) into a boolean to reflect how it is used.

libeos-payg/efi.c
```diff
@@ -544,15 +544,11 @@ efivarfs_read (const char  *name,
...
+  g_autofree char *name = full_efi_name (GLOBAL_VARIABLE_GUID, "SecureBoot");
   g_autofree unsigned char *content = NULL;
   int size;
 
-  /* In test mode let's pretend secure boot is enabled */
-  if (test_mode)
-    return TRUE;
-
-  content = efivarfs_read (tname, &size, NULL);
+  content = eospayg_efi_var_read_fullname (name, -1, &size, NULL);
```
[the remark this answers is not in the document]
wjt: See below, this comment is invalidated by the next change – it would be less confusing not to do that.

libeos-payg/efi.c
```diff
@@ -544,15 +544,13 @@ efivarfs_read (const char  *name,
 gboolean
 eospayg_efi_secureboot_active (void)
 {
-  g_autofree char *name = full_efi_name (GLOBAL_VARIABLE_GUID, "SecureBoot");
-  g_autofree unsigned char *content = NULL;
-  int size;
+  /* In test mode let's pretend secure boot is enabled */
```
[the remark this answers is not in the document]
starnight: ah!  Thanks for the catch.  My brain was stuck.

wjt (approved):

PR #122, opened 2024-07-09
Report GError when EFI functions fail

wjt:
Previously, many functions to read and manipulate EFI variables would return FALSE or NULL on error, but there was no reliable way to determine the cause of the error. Some error paths would set `errno`, but not all.

Add a `GError **` parameter to these functions, and set it whenever FALSE or NULL is returned.

In addition, extend `eospayg_efi_var_read()` to check whether the variable contents match the caller's expected size, which almost all callers would previously do. This way we only need to write the logic to set the GError in this case once, here, and the callers' error handling is simplified.

I did not add GError ** parameters to the various miscellaneous accessors.

This PR requires the out-of-tree plugins to be updated in lockstep because I chose not to preserve the API & ABI.

https://phabricator.endlessm.com/T35459

### DIFF
--- a/libeos-payg/efi.c
+++ b/libeos-payg/efi.c
@@ -53,7 +53,8 @@ struct efi_ops {
                      int          size,
                      gboolean     allow_overwrite,
                      GError     **error);
-  gboolean (*delete) (const char *name);
+  gboolean (*delete) (const char  *name,
+                      GError     **error);
   void (*list_rewind) (void);
   const char *(*list_next) (void);
   gboolean (*clear) (void);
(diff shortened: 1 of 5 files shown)


libeos-payg/efi.c
```diff
@@ -62,25 +63,26 @@ struct efi_ops {
...
 {
   unsigned int flags;
   int ret = 0;
   glnx_autofd int fd = -1;
 
   if (efi_fd == -1)
-    return FALSE;
+    return glnx_throw (error, "EFI ops not initialized");
 
   fd = openat (efi_fd, name, O_RDONLY);
```
dbnicholson, line 76: Am I wrong or should this throw here if the file isn't opened? Otherwise the `ioctl`s below would fail with `EBADF`, which I believe gets mapped to `G_IO_ERROR_FAILED`. Not that the callers are currently distinguishing errors, but IMO they should.

dbnicholson (approved):
Lovely. I had one nitpick on the last commit, but this is a very nice improvement. I'll leave it to you to merge since it requires coordination.

libeos-payg/efi.c
wjt: Good catch! I spotted & fixed this issue in _read but not here.

libeos-payg/efi.c
```diff
@@ -196,7 +198,7 @@ efivarfs_write (const char  *name,
 
   /* It may not exist, and this will fail harmlessly */
   if (allow_overwrite)
-    clear_immutable (name);
+    clear_immutable (name, NULL);
```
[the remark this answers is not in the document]
wjt: I also wondered about changing `clear_immutable` to operate on a file descriptor rather than opening it, changing its attribute, and closing it; only for this caller to open it again. But the test coverage here is nonexistent – it's not easy to mock efivarfs! –  so I preferred to leave it. The attempt to write will fail with (presumably) a meaningful error message anyway.

dbnicholson (approved):
LGTM

starnight (approved):
Nice!!

PR #123, opened 2024-07-09
Re-enable TAP parsing for C tests

wjt:
This reverts commit 2f1171b16ecd5e9d4e042e8faac1eae452a6340c, which reverted part of commit 15f0d9319f29c56cdd345b67892013786084921b.

We now have a newer GLib which emits the "TAP version XX" header, so let's try again to tell Meson that the C tests emit TAP output, and bump the GLib dependency to the version that added that header.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -9,7 +9,7 @@ Build-Depends:
  dh-python,
  dracut,
  gtk-doc-tools,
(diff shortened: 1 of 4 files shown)


dbnicholson (approved):
Looks like a winner!

PR #127, opened 2026-03-09
Introduce gen paygid tool

starnight:
Introduce generating new PAYG IDs & Keys tool as gen_paygid.py and its test cases. The tool will generate PAYG ID & private keys with Angaza v2 API endpoint. And, it will be installed as `<bindir>/eos-gen-paygid`.

Then, package it into the new eos-payg-admin package.

https://app.asana.com/1/1203676861188277/project/1211949321657103/task/1213224911599528


dsd:
I haven't looked at this code in detail, nor have I considered this suggestion in detail.
But given that this is an infrequent action performed by just 1-2 people, I am wondering if it would make sense to add this as a tool to the eos-admin container.
When it is run from that container context, you can directly pull the secrets from vault. The payg-payments python scripts show how this can be done. What do you think?

starnight:
That might work.  Will try!

starnight:
Suppressed by https://github.com/endlessm/eos-administration/pull/1558

